PR #101, opened 2026-05-30
feat(dogfood-ledger): canonical dogfood-cycle counter (Phase 1 — primitive)

vek-emilsonmoraes:
### Context
The dogfooding-mandate gates promotion on **≥2 real cycles**, but the count lived only as **changelog prose** across the ecosystem (ADR-017 R6, auto-pilot MASTER-PLAN, GIT-PR-AS-DEBATE, per-skill changelogs) — *unmeasurable = theater*. The operator flagged it directly: cycles **already happened across many sessions**, they were just never **counted**. This PR adds the structured, `jq`-countable, AAIF SSOT for dogfood-cycle counting, and lands it in MAOS (the canonical agentic-OS that owns the mandate) rather than any product repo.

### What's in this PR (Phase 1)
- **`bin/dogfood-mark`** — append-only writer. Idempotent (no dup of identical event), atomic `mkdir`-lock, **anti-theater gate** (`complete` REQUIRES `--ratified` + ≥1 `--evidence`, else refuse), `--dry-run`. Cloned from ASH `ash-decide` + `ash-stop-fallback`.
- **`bin/dogfood-tally`** — read-only **counting authority** (SRP). Reduces the event log by `(tool,cycle_id)` to latest-by-`ts`, counts COMPLETE/IN-PROGRESS/ABANDONED per tool, renders the gate verdict. Cloned from ASH `ash-decisions` + `ash-walkthrough` aggregator.
- **`docs/dogfood-cycle-ledger-spec.md`** + **`docs/adrs/ADR-005`** — spec + decision (alternatives rejected: per-skill island ledger / changelog-prose / product-repo / new-DB).
- **`skills/dogfood-ledger/SKILL.md`** — thin ergonomic wrapper (engine is vendor-neutral bash → Cursor/Copilot/Codex/Gemini too).

Ledger = user-scope `~/.claude/audit/dogfood-cycles.jsonl` (sees every project/session/vendor — portability). AAIF bash 3.2, jq-only, Layer-Purity clean.

### Acceptance / verification
- ✅ **11/11 sandbox tests pass under bash 3.2.57** (mark · idempotent no-op · anti-theater refuse ×2 exit 1 · in-progress→complete transition · dry-run · reduce-latest dedup · gate ELIGIBLE/NOT · slug-validation · empty-ledger exit 2).
- ✅ gitleaks clean · Layer-Purity clean (zero product-repo strings).
- ✅ **Recursive dogfood**: this mechanism's own cycle (`dogfood-ledger/001`) + the in-flight `debate-converge/001` are recorded in the real ledger with evidence — the counter is non-empty.

### Deliberately deferred (honest, documented — NOT in this PR)
- **Phase 2**: `--backfill` automation (scan transcripts/changelogs/ASH-journals for historical cycles, evidence-based) — spec/skill mark it `Phase 2 — planned`; until then, cycles are recorded via direct evidence-based `dogfood-mark` calls.
- **Phase 3**: portability install-script (PATH-accessible cross-project) · ASH `ash-walkthrough --dogfood-cycles` consumer · Sentinel `session_index` integration · migrate per-skill `DOGFOOD-LEDGER.md` → this SSOT.

### Out of scope
Graph/inter-relation (delegated to CPT §9 + `gsd-graphify`) · LGPD 5y out-of-band retention sink (separate ADR; `~/.claude/audit/` is local-only).

### DIFF
--- a/bin/dogfood-mark
+++ b/bin/dogfood-mark
@@ -172,7 +172,7 @@ while ! mkdir "$LOCK_DIR" 2>/dev/null; do
       continue
     fi
   fi
-  i=$((i + 1)); [ "$i" -ge 50 ] && { echo "could not acquire ledger lock (held >5s; if a process was SIGKILLed mid-write, remove $LOCK_DIR)" >&2; exit 2; }
+  i=$((i + 1)); [ "$i" -ge 50 ] && { printf 'could not acquire ledger lock (held >5s; if a process was SIGKILLed mid-write, remove %s)\n' "$LOCK_DIR" >&2; exit 2; }
   sleep 0.1
 done
 trap 'rm -f "$LOCK_DIR/ts" 2>/dev/null; rmdir "$LOCK_DIR" 2>/dev/null || true' INT TERM EXIT
--- a/bin/dogfood-mark
+++ b/bin/dogfood-mark
@@ -57,17 +57,21 @@ TOOL=""; CYCLE=""; STATUS="in-progress"; RATIFIED="false"
 SID="${CLAUDE_CODE_SESSION_ID:-}"; PROJECT=""; VENDOR=""; NOTE=""; DRYRUN=0
 EVID_RAW=""
 
+# Guard a value-taking flag: ensure a value follows before `shift 2` (else set -e would
+# abort on a short `shift` without a clear usage error). Call as: need_val --flag "$#".
+need_val() { [ "$2" -ge 2 ] || { echo "$1 requires a value" >&2; exit 1; }; }
+
 while [ $# -gt 0 ]; do
   case "$1" in
     --help|-h) usage ;;
-    --status) STATUS="${2:-}"; shift 2 ;;
+    --status) need_val --status "$#"; STATUS="$2"; shift 2 ;;
     --ratified) RATIFIED="true"; shift ;;
-    --evidence) EVID_RAW="${EVID_RAW}${2:-}
+    --evidence) need_val --evidence "$#"; EVID_RAW="${EVID_RAW}$2
 "; shift 2 ;;
-    --session) SID="${2:-}"; shift 2 ;;
-    --project) PROJECT="${2:-}"; shift 2 ;;
-    --vendor) VENDOR="${2:-}"; shift 2 ;;
-    --note) NOTE="${2:-}"; shift 2 ;;
+    --session) need_val --session "$#"; SID="$2"; shift 2 ;;
+    --project) need_val --project "$#"; PROJECT="$2"; shift 2 ;;
+    --vendor) need_val --vendor "$#"; VENDOR="$2"; shift 2 ;;
+    --note) need_val --note "$#"; NOTE="$2"; shift 2 ;;
     --dry-run) DRYRUN=1; shift ;;
     -*) echo "Unknown option: $1" >&2; exit 1 ;;
     *)
@@ -154,15 +158,21 @@ LOCK_DIR="$LEDGER_DIR/.lock-dogfood"
 STALE_SECS=30
 i=0
 while ! mkdir "$LOCK_DIR" 2>/dev/null; do
+  # Stale-lock reclaim: a writer killed by an uncatchable SIGKILL leaves the lock behind.
+  # Read its epoch marker (numeric-guarded against empty/corrupt); if older than STALE_SECS,
+  # reclaim ATOMICALLY via `mv` — rename is POSIX-atomic, so only ONE concurrent process wins
+  # the rename (the losers get ENOENT), eliminating the check-then-remove TOCTOU. The winner
+  # discards the moved dir; everyone re-races mkdir so exactly one holder remains.
   if [ -f "$LOCK_DIR/ts" ]; then
-    age=$(( $(date +%s 2>/dev/null || echo 0) - $(cat "$LOCK_DIR/ts" 2>/dev/null || echo 0) ))
-    if [ "$age" -gt "$STALE_SECS" ]; then
-      rm -f "$LOCK_DIR/ts" 2>/dev/null || true
-      rmdir "$LOCK_DIR" 2>/dev/null || true
+    lock_ts=$(cat "$LOCK_DIR/ts" 2>/dev/null || echo 0); case "$lock_ts" in ''|*[!0-9]*) lock_ts=0 ;; esac
+    now=$(date +%s 2>/dev/null || echo 0);              case "$now"     in ''|*[!0-9]*) now=0 ;; esac
+    if [ "$((now - lock_ts))" -gt "$STALE_SECS" ]; then
+      dead="$LOCK_DIR.stale.$$"
+      if mv "$LOCK_DIR" "$dead" 2>/dev/null; then rm -rf "$dead" 2>/dev/null || true; fi
       continue
     fi
   fi
-  i=$((i + 1)); [ "$i" -ge 50 ] && { echo "could not acquire ledger lock (held >5s)" >&2; exit 2; }
+  i=$((i + 1)); [ "$i" -ge 50 ] && { echo "could not acquire ledger lock (held >5s; if a process was SIGKILLed mid-write, remove $LOCK_DIR)" >&2; exit 2; }
   sleep 0.1
 done
 trap 'rm -f "$LOCK_DIR/ts" 2>/dev/null; rmdir "$LOCK_DIR" 2>/dev/null || true' INT TERM EXIT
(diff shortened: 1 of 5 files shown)


vek-emilsonmoraes:
@amazon-q-developer @qodo-code-review — all findings addressed in `ca0ad2c`.

| # | Finding | Verdict | Fix |
|---|---------|---------|-----|
| AQ-1 | `dogfood-tally` crashes on corrupted JSONL (`set -e`) | ✅ fixed (root-cause) | Resilient `jq -R 'fromjson? // empty'` parse — skips malformed/partial lines + emits a skipped-line stderr diagnostic. Applied to **every** ledger read-path (tally **and** mark's idempotency check), so it also neutralizes the AQ-3 "partial write remains" concern. |
| AQ-2 | `date` failure → malformed TS | ✅ fixed | `TS=$(date … 2>/dev/null) \|\| exit 2` + non-empty guard — no malformed event is ever written. |
| AQ-3 | Lock: stale on SIGKILL + trap only on EXIT | ✅ fixed | `trap … INT TERM EXIT` (catchable signals) + epoch-marker stale-lock reclaim (>30s) so a SIGKILL'd holder no longer wedges future writers. |
| Qodo-1 | Idempotency race window (check outside lock) | ✅ fixed | Authoritative idempotency re-check moved **inside** the `mkdir`-lock (closes the TOCTOU dup window); the pre-lock check is kept as an advisory fast-path only. |
| Qodo-2 | `group_by` on unsorted array → wrong counts | ⚖️ partial | The **grouping-correctness** claim is a respectful **false positive** — jq `group_by` sorts internally before grouping, so append-order interleaving is handled correctly (`echo '[3,1,2,1]' \| jq -c 'group_by(.)'` → `[[1,1],[2],[3]]`). The **real** gap your note surfaced was the `max_by(.ts)` tie-break at second-granularity `ts`: ✅ fixed with `max_by([.ts, .__idx])` (later-appended line wins on a tie) + array group-key `[.tool,.cycle_id]`. |

**Verification:** 22/22 sandbox tests pass under bash 3.2.57 (incl. new cases: malformed-line resilience ×2, date-guard, tie-breaker determinism, array-key no-collision) · gitleaks clean · plugin validation PASSED (0 errors/0 warnings) · real canonical ledger re-parses with no regression.

Copilot generated no comments; CodeRabbit was rate-limited (no findings). Could you please re-review `ca0ad2c`?

vek-emilsonmoraes:
@coderabbitai review

vek-emilsonmoraes:
@amazon-q-developer @coderabbitai — re-review findings addressed in `0c24ea2` (PDCA iter-2).

| Finding | Verdict | Fix |
|---------|---------|-----|
| **amazon-q** `mark:163` 🛑 stale-lock reclaim TOCTOU | ✅ fixed | Reclaim is now **atomic via `mv`** — rename is POSIX-atomic, so only ONE concurrent process wins the rename (losers get `ENOENT`); the winner discards the moved dir and everyone re-races `mkdir`, so exactly one holder remains. This closes the check-then-remove window more cleanly than `rm -f`-gating (which still returns 0 on a missing file). |
| **CodeRabbit** `mark:170` 🟠 harden stale-marker parse | ✅ fixed | Marker `case` numeric-guarded (`''\|*[!0-9]*) → 0`) so a corrupt/empty marker can't abort the arithmetic under `set -e`; folded into the reclaim above. |
| **CodeRabbit** `mark:70` 🟡 validate value flags before `shift 2` | ✅ fixed | `need_val` helper guards `--status/--evidence/--session/--project/--vendor/--note` → missing value prints a clear error + `exit 1` instead of a `set -e` short-`shift` abort. |
| **CodeRabbit** `tally:40` 🟡 guard `--gate` before shift | ✅ fixed | Same `need_val` helper in `dogfood-tally`. |
| **CodeRabbit** `spec:56` 🟡 don't list `--backfill` as supported | ✅ fixed | §5 table now annotates `--backfill` as **Phase 2, planned; not yet implemented** (consistent with §6's banner). |

amazon-q's original 3 stop-signs (partial-write / date / JSONL) are resolved — GitHub marked them **OUTDATED** after `ca0ad2c`.

**Verification:** 28/28 sandbox tests pass under bash 3.2.57 (+6 iter-2: arg-guard ×2, atomic stale-lock reclaim ×3, corrupt-marker numeric-guard ×1) · gitleaks clean · plugin validation PASSED (0/0) · real canonical ledger re-parses with no regression.

Could you please re-review `0c24ea2`?

vek-emilsonmoraes:
@coderabbitai review

vek-emilsonmoraes:
@amazon-q-developer — iter-3 re-review: the 3 new 🛑 findings are **false positives**. Verified empirically (not asserted), one micro-hardening adopted in `ed04f6b`.

### 🛑 `mark:175` "Command Injection via `DOGFOOD_LEDGER_DIR`" — **FALSE POSITIVE**
Bash does **not** re-evaluate command-substitution that arrives via *variable expansion*. `$(...)` inside the **value** of `$LOCK_DIR` is printed literally by `echo "$LOCK_DIR"` — it is never executed (command substitution only fires on literal `$(...)` in source text, not on text produced by expansion). Proof:
```console
$ X='$(touch /tmp/INJECTED_PROOF)'; echo "remove $X"; ls /tmp/INJECTED_PROOF
remove $(touch /tmp/INJECTED_PROOF)
ls: /tmp/INJECTED_PROOF: No such file or directory   # ← NOT executed
```
There is no CWE-78 here. **However**, I adopted your `printf '%s'` suggestion in `ed04f6b` anyway — it is a genuine micro-hardening (robust against a `LOCK_DIR` that starts with `-` or contains backslashes, which `echo` could misinterpret), independent of the injection framing.

### 🛑 `mark:186` "Trap cleanup race between lines 185–186" — **FALSE POSITIVE**
The lock is **held** throughout 183–188: `mkdir "$LOCK_DIR"` already succeeded and the `EXIT` trap (line 178) only releases it when the process exits — which happens **after** the append on line 188 (or after the `exit 0` on line 185). A concurrent writer cannot `mkdir` the still-existing dir, so there is no 185–186 window. No change needed.

### 🛑 `tally:90` "Empty-group crash on `.[0].tool`" — **FALSE POSITIVE**
`group_by` never yields an empty sub-group — it partitions existing elements. The `select(.tool and .cycle_id and .ts)` filter runs on the **flat array before** the first `group_by`, so it removes events, it does not create empty groups. If *every* event is filtered out, the array is `[]` → both `group_by` calls return `[]` → `map(…)` over `[]` never executes the lambda, so `.[0].tool` is never reached. Proof:
```console
$ printf '%s\n' '{"foo":1}' '{"bar":2}' > led.jsonl   # valid JSON, invalid schema
$ DOGFOOD_LEDGER_DIR=. dogfood-tally --json
[]                                                     # exit 0, no crash
```
Adding `map(select(length>0) | …)` would be dead code defending an unreachable state, so I left it out (per anti-theater/KISS).

---

**Verification of `ed04f6b`:** 28/28 sandbox tests pass (bash 3.2.57) · gitleaks clean · plugin validation PASSED · real ledger no regression.

**Convergence summary:** all genuine findings across iter-1 (`ca0ad2c`) + iter-2 (`0c24ea2`) were fixed; iter-3 surfaced only false positives (rebutted above with proof). Status checks are 11/11 green (`mergeable: CLEAN`). Per the diminishing-returns convergence rule, this PR is now **held for the maintainer's manual merge decision** (author == maintainer; dogfood-gate is a maintainer-policy call).

Tracked in **VKS-2059**.

vek-emilsonmoraes:
@coderabbitai

PDCA iter-4 (`febf549`) — verified all 4+1 actionable against current HEAD; addressed the still-valid ones (per your "fix only still-valid" instruction):

| Finding | File | Verdict | Fix |
|---|---|---|---|
| 1 (Major / quick-win) | `skills/dogfood-ledger/SKILL.md` | **still-valid → fixed** | Restructured to repo skill contract: `Why/Use/Rules` → `Purpose` / `When to Use` / `Trigger Phrases` / `Protocol Rules`. Content preserved; plugin validation PASSED (0 errors/0 warnings). |
| 2 | `bin/dogfood-mark` lock-ts | already addressed (iter-2) | `lock_ts` sanitized to `0` on empty/non-numeric (`case … in ''\|*[!0-9]*)`). |
| 3 | `bin/dogfood-mark` arg-guard | **still-valid → fixed** | `need_val` now also rejects a flag-prefixed value; value passed as `${2:-}` to stay `set -u`-safe on missing value. |
| 4 | `bin/dogfood-tally --gate` | already addressed | `need_val` guard + integer-validate (`case "$GATE" in *[!0-9]*\|"")`) — rejects flag-prefix too. |
| 5 | `docs/dogfood-cycle-ledger-spec.md` | **still-valid → fixed** | Dropped `--backfill` from the dogfood-mark guarantees cell; kept §6 Phase-2 explanation (no longer advertises an unimplemented flag). |

amazon-q's 3× `:stop_sign:` were previously proven false-positive (iter-3, empirical). Could you please re-review?

PR #122, opened 2026-06-09
feat(postflight): scorecard gallery — deterministic 7-model renderer + templates

vek-emilsonmoraes:
**[PR-as-Enhancement-Prompt]** Candidate end-of-action output formats for `postflight` P3 HANDOFF + the `end-of-action-briefing` §4.1 legend.

**ONE renderer · ONE param schema · 7 models** — `bin/scorecard.py`. Self-calculates every bar/%/count/tally/sparkline/burndown + (`--auto-git`) git/PR facts; the AI agent supplies only the atoms (verdict, per-item state/label/note/confidence, what's-left) as a JSON params blob. 5-state legend 🔴🟡🟠🟢🔵 (icon+word+machine-token), pulse %green, `NO_COLOR`/`--no-color`, stdlib-only.

Models: 1 Cockpit · 2 Traffic-Light Strip · 3 KPI Tiles · 4 Burndown Ledger · 5 Kanban Lanes · 6 Telemetry/JSON-RPC · 7 Executive One-Liner. Scoring + recommendation in `skills/postflight/scorecards/GALLERY.md`.

**Additive** — does NOT yet change postflight's default output; the operator picks which model(s) become the default in a follow-up. Verified: `--all --demo` (7 render), `--auto-git`, `--params` paths; Layer-Purity 0 violations; gitleaks clean.

Files: `bin/scorecard.py` · `skills/postflight/scorecards/{GALLERY.md,sample-params.json}`

### DIFF
--- a/bin/scorecard.py
+++ b/bin/scorecard.py
@@ -1,0 +1,469 @@
+#!/usr/bin/env python3
+"""
+postflight scorecard renderer — deterministic end-of-action status cards.
+
+ONE renderer · ONE param schema · 7 layout MODELS. Design contract:
+  - DETERMINISTIC: same params -> same output (no clock unless passed in).
+  - SELF-CALCULATING: auto-derives git/PR facts + the autonomy pulse + bars
+    from numbers, so an AI agent only supplies what cannot be computed.
+  - AI-PARAMETERIZED: the qualitative items (verdict, checklist, notes,
+    confidence, what's-left) arrive as a JSON params blob (file/stdin).
+  - HUMAN-GLANCEABLE: 5-state colour legend (🔴🟡🟠🟢🔵) + bars + tally.
+  - PORTABLE: stdlib only; NO_COLOR / --no-color honoured; emoji carry colour
+    so it renders in any terminal AND in GitHub/markdown.
+
+Usage:
+  scorecard.py --model N [--params FILE|-] [--demo] [--auto-git] [--repo DIR]
+               [--no-color] [--all]
+  echo '<json>' | scorecard.py --model 2
+  scorecard.py --all --demo            # render every model with demo data
+
+Param schema (all keys optional; sane fallbacks):
+  {
+    "session":  {"title","id","date","time","duration"},
+    "verdict":  {"state":"done|warn|blocked|wip", "label":"DONE · green · merged"},
+    "autonomy": {"green":N,"blue":N,"orange":N,"red":N},      # pulse tally
+    "vitals":   [{"label","icon","pct":0-100,"note"}],
+    "checklist":[{"state":"green|blue|orange|yellow|red","label","note","confidence":0-100|null}],
+    "whats_left":[{"state":"done|orange|red|yellow","text"}]
+  }
+"""
+import argparse, json, os, sys, subprocess, datetime, re
+
+# ── 5-state legend (icon · human-word · machine-token · ansi-256) ─────────────
+STATE = {
+    "red":    ("🔴", "not-started", "not_started",  196),
+    "yellow": ("🟡", "in-progress", "in_progress",  226),
+    "orange": ("🟠", "need-HITL",   "need_hitl",    208),
+    "green":  ("🟢", "done-IA",     "done_agentic",  46),
+    "blue":   ("🔵", "done-human",  "done_human",    39),
+}
+ALIAS = {"done": "green", "agentic": "green", "human": "blue", "hitl": "orange",
+         "doing": "yellow", "wip": "yellow", "todo": "red", "blocked": "red",
+         "warn": "orange"}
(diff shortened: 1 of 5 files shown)


vek-emilsonmoraes:
### 🛠️ Added: fix `spawn-continuation.sh` (postflight P3.5 launcher)

Commit `f9bb189` — atomic fix for 3 empirically-diagnosed bugs surfaced while dogfooding the spawn (the seeded continuation session reported "spawned" but `tmux attach` → *no sessions* and the handoff `claude --session-id <id>` → *"Session ID already in use"*).

| # | Bug | Fix |
|---|---|---|
| 1 | **cmux fake-spawn** — `cmux)` case ran `cmux claude … &` then *unconditionally* claimed `spawned (cmux)` + `exit 0`. cmux has no "run a detached command" primitive (its CLI opens paths/URLs). | auto-detect prefers **tmux** (only verifiable detacher); forced `MAOS_SPAWN_LAUNCHER=cmux` now no-ops → honest `register+print` (no fake "spawned"). |
| 2 | **session-id collision** — printed re-entry command hardcoded `--session-id <uuid>`, which **creates** a session; re-running (or running after the tmux path already created it) → *"already in use"*. | re-entry hints + JSON `attach` field now use **`claude --resume <uuid>`** (start-once via `--session-id`, re-enter via `--resume`). |
| 3 | **doc** advertised a non-existent `(or via cmux)` attach path. | usage now documents `claude --resume <uuid>` re-entry + the `--session-id` vs `--resume` distinction. |

**Root cause (Mente Tomé)**: child session `77ceb924` had a 72KB transcript (tmux spawn *did* run) but the pane had exited — the correct re-entry was `--resume` all along, not a fresh `--session-id`.

**Verified**: `bash -n` ✓ · `--dry-run --force` → `launcher=tmux` + `attach=claude --resume` ✓ · forced-cmux → `status=registered` (no fake spawn) ✓ · gitleaks clean · GAAS pre-commit/pre-push pass.

> Scope: bug-fix only (no redesign). The 7-model scorecard gallery + design-polish pass remain the primary PR deliverables.

vek-emilsonmoraes:
## ✅ PDCA convergence — all reviewer findings addressed (`331d150`)

Each finding was **verified against the current code** before acting (the file shifted after the `tickets` commit `1de0260`, so reviewer line numbers were from an earlier revision). 2 amazon-q diagnoses turned out to be false-positives — flagged honestly below.

### 🔴 @coderabbitai (blocking — `CHANGES_REQUESTED`)
| Finding | Fix | Status |
|---|---|---|
| `GALLERY.md` uppercase violates lowercase-hyphenated `skills/**` convention | `git mv GALLERY.md → gallery.md` (history preserved); grep confirms **zero** lingering refs | ✅ |
| MD031 — fenced blocks need surrounding blank lines | blank line added after both closing fences (json schema block + bash usage block); `markdownlint-cli2` now reports **0 MD031** | ✅ |
| stdin example uses positional `-` (CLI has no positional stdin) | `--model 2 -` → `--model 2 --params -` (verified against `load_params()`) | ✅ |

### 🟡 @amazon-q-developer
| # | Finding | Fix | Status |
|---|---|---|---|
| 1 | `_run()` broad `except` masks failures (CWE-755) | narrowed to `(subprocess.SubprocessError, OSError)` — git/gh still degrade gracefully, but programming bugs now surface; also narrowed the twin `json.loads` except | ✅ |
| 2 | `load_params()` file-handle leak on read error | `with open(...) as f: raw = f.read()` | ✅ |
| 3 | `model_5()` redundant `setdefault` "loses items" | **⚠️ false-positive** — `setdefault` returns the *existing* list, so no items were ever lost; the extra line was merely **dead code** and is now removed | ✅ (cleanup) |
| 4 | `model_6()` line crash risk | **⚠️ partial false-positive** — that line was already guarded (`if k in a … if a else ""`); the *real* crash was `a['green']`/`a['blue']` on partial params → fixed at the root (see qodo below). Bonus: `spark` was computed but **never rendered** → wired into the telemetry output, completing the advertised sparkline | ✅ |

### 🟢 @qodo-code-review (root cause)
| Finding | Fix | Status |
|---|---|---|
| "all keys optional" but renderers index nested keys → `KeyError` on partial JSON | **`enrich()` now deep-normalizes** the whole blob (autonomy→4 int keys; checklist/whats_left/vitals/tickets→dicts with safe `state`/`label`/`note`/`text`; new `_norm_state()` for unknown/aliased states). One fix makes **all 7 models** crash-proof — the KeyError class is gone | ✅ |

### Tests (all pass)
`py_compile` · all-7-models `--demo` · **partial-params crash test** (unknown state + missing keys, the qodo repro) · malformed-types (autonomy as string, etc.) · sparkline render · stdin `--params -`. `gitleaks` clean. `ruff` clean on all changed lines (pre-existing `E702`/`F401` in the author's compact style are out of scope for this convergence cycle).

Could you please re-review? 🙏

@coderabbitai review

PR #258, opened 2026-07-14
feat(system-health-responder): EKO-90 round-4 — threshold false-positive fixes (v1.6.0)

ekson73:
## [PR-as-Enhancement-Prompt] EKO-90 round-4 — threshold false-positive fixes

**Context** — A `/quiesce --scope [linear:EKO-90] --now` OODA pass after the round-3 merge (#257). Two false-positive classes, both **empirically observed this session**, made the health-contract mis-classify — which would make the end-of-action responder engage spuriously. A false `crit`/`warn` isn't cosmetic: it erodes the trust needed to let the auto-heal run unattended, so this is squarely an **enhance-autonomy** fix. Collector `v1.3.2 → v1.4.0` · skill `v1.5.1 → v1.6.0`.

**Objectives** — kill the two false-positives *without* weakening real-risk detection; keep it Gordian (2 threshold cores, no scope-creep).

| Gap (empirical) | Root cause | Fix |
|---|---|---|
| CPU `crit` on a transient burst (`load1=27.33/12c`, settled to `warn` in ~4 min) | both warn+crit keyed on `load1` — the *spikiest* (1-min) metric | `cpu_load_status()`: **WARN on load1** (early), **CRIT on load5** (sustained). A 1-min burst warns, never crits. |
| XProtect chronic `warn` (5347 @ 47d) even though 5347 **is** Apple's latest | age-alone conflates "stale" with "Apple hasn't shipped" | `xprotect_status()`: gate on the **auto-update channel** (~20ms `softwareupdate --schedule`, no-sudo). ON ⇒ self-healing (caps at `warn` past 60d, never crit); OFF ⇒ full 30/60 escalation (the real risk). |

Both extracted as **pure functions** (mirroring `compute_burndown`) + **test entrypoints** (`--cpu-load-status` / `--xprotect-status`) so they're exercised, not eyeballed. New drill-down fields: `load5`, `load5_ratio`, `auto_update`.

**Acceptance**
- [x] `bin/threshold-test.sh` — 13 assertions (burst→warn-not-crit · sustained-load5→crit · auto-ON-47d→ok · auto-OFF-65d→crit · never-crit-while-self-healing)
- [x] Round-3 suites regress clean: `burndown-test.sh` 16/16 · `responder-burndown-test.sh` 5/5
- [x] Live collector run → contract `v1.4.0`, `secret_free:true`, gitleaks clean, contract now honestly `ok`
[...]

PR #259, opened 2026-07-14
feat(system-health-responder): EKO-90 round-5 — ncpu-aware top_consumer + honest XProtect + pid-recycle guard (v1.7.0)

ekson73:
## [PR-as-Enhancement-Prompt] EKO-90 round-5 — finish the ncpu-aware thesis + measurement honesty + safer autonomy

**Context** — A `/quiesce --scope [linear:EKO-90] --now` OODA pass after round-4 (#258). A skeptic gap-sweep (Mente Tomé) found that round-4 made the **load-aggregate** ncpu-aware but left the **per-process** leaf raw — a genuine, *currently-firing* false-crit — plus three adjacent honesty/safety gaps. The sweep also correctly ruled ~5 candidates *works-as-intended* or *low-payoff* (network-expansion, ENGAGE-aggressiveness, memory-signal), so this round is deliberately narrow. Collector `v1.4.1 → v1.5.0` · skill `v1.6.1 → v1.7.0`.

**Objectives** — kill the last CPU false-crit, make two signals *honest*, harden the one autonomous action — without weakening real-risk detection. Gordian: 4 cohesive S-effort fixes, no scope-creep.

| # | Gap (empirical) | Root cause | Fix |
|---|---|---|---|
| **1** | `top_consumer`/`runaway` **crit on 1.3 of 12 cores** (live `global=crit` while `load5` said fine — the branches *contradicted* each other) | warn+crit applied raw to `ps %cpu`, which is **per-core** (100% = one core), no ncpu normalization | `top_consumer_status()`: **WARN per-core** (renice-actionable signal survives) · **CRIT = `pct ≥ PROC_CPU_CRIT_RATIO(0.50)×ncpu×100`** (fraction of *total* capacity). ncpu=1 fail-safe. Live `crit→warn`. Auto-resolves the old renice+kill double-count. |
| **5** | XProtect age measured the **wrong artifact** (bundle-mtime 47d) vs the authoritative install (Jun-3 = 41d) | modern macOS ships signature updates via `XProtectUpdateService`, never touching `XProtect.bundle` | prefer `/usr/bin/xprotect version` (no-sudo, real version+date); `source` leaf = `xprotect-cli\|bundle-mtime`; bundle-mtime is the fallback for older macOS. `medição real`. |
| **2** | responder could `renice` a **recycled pid** (contract ≤600s stale; denylist ran against the *stale* comm) | `try_renice()` validated pid-alive but not that the live proc still matches | re-read live `comm`, require it to still match the contract comm **before** acting; re-run the denylist on the *live* comm. Safer autonomy. |
| **6** | `process.counts` leaf hardcoded `status:"ok"` while its own `zombies` could drive the branch `crit` → root-fail drill-down landed on an `ok` leaf | hardcode + inline-computed zombie status (2×) | compute `ZOMBIE_ST` once (DRY); leaf reflects it — honors the DoD's "cada nível aponta o filho com falha". |

**Deferred (documented, not dropped)** — memory could use the native `kern.memorystatus_vm_pressure_level` signal (low payoff: memory is HITL-only + the current available% already adds back reclaimable classes).

**Acceptance**
- [x] `bin/threshold-test.sh` — **+7 `top_consumer_status`** assertions (130.7@12c→warn · 700@12c→crit · 600 boundary→crit · 85→warn · 50→ok · ncpu=1 fail-safe both ways) = **23**
- [x] Round-3/4 suites regress clean: `burndown-test.sh` 16 · `responder-burndown-test.sh` 5 → **44/44 total**
- [x] Live collector → `v1.5.0`, `secret_free:true`, `global crit→warn` (by correct classification), `xprotect.source:"xprotect-cli"` age 41d, gitleaks clean
- [x] **Real-risk preserved** — `700@12c → crit` (genuine multi-core hog) still fires; `auto-OFF 65d → crit` (round-4) intact
- [x] Anti-theater — fixes *what "healthy"/"stale" mean*, does not suppress; responder still renices any non-`ok` top_consumer

**Test plan** — `bin/threshold-test.sh && bin/burndown-test.sh && bin/responder-burndown-test.sh` (44/44).

**Risk / rollback** — threshold + read-source + a defensive skip-path only; reversible (`git revert`). Machine-local collector backed up. Responder dispositions unchanged (still gates purely on computed leaf status — DRY single-owner).

**Note on checks** — **Snyk is being discontinued** (operator-confirmed 2026-07-14); its quota-`error` statuses are not a gate. The repo's native suite (gitleaks · Trivy×2 · pip-audit · OpenSSF · governance · Lockfile) is the security gate.

**Cross-links** — EKO-90 (Linear, team EKO) · round-4 #258 (`b4f3485`) · collector `~/.local/bin/system-health-guardian.sh` v1.5.0 (live, validated).

---
> 🤖 agent-authored — Claude Opus 4.8 under operator `/quiesce round-5` directive 2026-07-14.

https://claude.ai/code/session_01EYY7MuHqYHm6fKNLjrrU2Z

### DIFF
--- a/skills/system-health-responder/SKILL.md
+++ b/skills/system-health-responder/SKILL.md
@@ -248,8 +248,11 @@ the unfinished half plus three adjacent honesty/safety gaps (collector **v1.4.1 
   autonomous action the responder takes. *(v1.7.1 PDCA per CodeRabbit PR#259 #131/#138 — the initial
   either-contains-substring match was too loose: a recycled `foo-helper` would match a contract `foo`, and an
   empty contract comm would match anything, either authorizing a wrong-proc renice; replaced with exact
-  equality + start-identity re-check. `renice` is fully reversible + denylist-guarded + re-sampled every 600s,
-  so fail-closed is sound.)*
+  equality + start-identity re-check. A follow-up review added two more: **@145** — `denied()` now
+  basename-normalizes so the whole-word `op` (1Password CLI) rule protects a PATH comm (`/usr/local/bin/op`),
+  not just a bare `op`; **@157** — the original niceness `cur` is now sampled AFTER the start-identity is bound
+  (a recycle can no longer pair a stale nice value with a live identity). `renice` is fully reversible +
+  denylist-guarded + re-sampled every 600s, so fail-closed is sound.)*
 - **#6 drill-down leaf-status consistency** — the `process.counts` leaf carried a hardcoded `status:"ok"`
   while its own `zombies` value could drive the branch to `crit`, so a `root_fail → leaf` walker landed on an
   `ok`-labeled leaf. Now the leaf reflects the (DRY, computed-once) `zombie` status — honoring the contract's
@@ -258,7 +261,8 @@ the unfinished half plus three adjacent honesty/safety gaps (collector **v1.4.1 
   signal (low payoff — memory is HITL-only + the current available% already adds back reclaimable classes).
 - **Tested**: `bin/threshold-test.sh` — **+7 `top_consumer_status`** (130.7@12c→warn · 700@12c→crit ·
   ncpu=1 fail-safe · …) **+8 `comm_match` recycle-guard** (foo-helper≠foo · empty≠anything ·
-  case/basename-normalize · …) assertions = **31**; round-3 suites regress clean (16+5) = **52/52**.
+  case/basename-normalize · …) **+6 `denied` basename-normalize** (/path/op→denied · top→allowed ·
+  1Password-path→denied · …) assertions = **37**; round-3 suites regress clean (16+5) = **58/58**.
 
 ## Composition (reuse, not reinvent — Strata)
 
--- a/skills/system-health-responder/SKILL.md
+++ b/skills/system-health-responder/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: system-health-responder
 description: End-of-action reflex that reads the system-health contract, engage-locks, Eisenhower-ranks the warnings, does MODERATE non-destructive auto-heal (autonomous reversible renice of a clear cpu runaway + `uv cache prune` producer-hygiene), responds to the new `agentic_tools` branch (`claude doctor` runtime health + cache-producer pressure), reads the reclaim-aware `burn_down` disk forecast (observe-only leading indicator → seed+notify before crisis), and escalate-seeds the HITL residue (security · malware · kill · disk→disk-guardian · no-source-fix offender containment). EKO-90 part-2 — the responder half of the health suite.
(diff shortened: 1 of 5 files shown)


ekson73:
@coderabbitai review

**@qodo-code-review** — acknowledged: the round-5 #2 pid-recycle guard does add one `ps -o comm=` probe + a skip-branch. That cost is **intended** — it's the price of not renicing a process that recycled into the top pid since the ≤600s-old contract sample (and the guard re-runs the denylist on the *live* comm, so a recycle-into-protected proc is refused). One cheap probe for a safer autonomous action is the correct trade for the responder's only self-acting lever. Non-blocking.

ekson73:
@coderabbitai — both Major findings on the pid-recycle guard fixed in `31908a9`:

**#131 (exact command match):** removed the either-contains-substring match. New `comm_match()` normalizes (basename + lowercase + strip-space) and requires **both sides non-empty + exactly equal**. A recycled `foo-helper` no longer matches contract `foo`, and an empty comm matches nothing — both now fail-closed (skip, no renice). Regression-locked with 8 new `bin/threshold-test.sh` assertions pinning exactly these cases (via a `--comm-match` test hook).

**#138 (TOCTOU):** bind a stable start-identity (`ps -o lstart=`) right after the comm match and **re-verify it immediately before `renice`** — drift or vanish ⇒ fail-closed skip. Kept responder-local (no contract change) as the proportionate close-out: `renice` is fully reversible (reverts.log), the denylist is re-checked on the *live* comm, and the collector re-samples every 600s so a real runaway re-surfaces next cycle — fail-closed defers at most one cycle rather than ever renicing the wrong process.

Suite 31/31 (52/52 with round-3 suites). gitleaks clean. Responder-only — collector untouched.

vek-emilsonmoraes:
@coderabbitai — both follow-up findings fixed in `bfe52e5`:

**@145 (denylist basename-normalize):** `denied()` now basename-lowercases its input before matching, consistent with `comm_match()`. A PATH comm `/usr/local/bin/op` normalizes to `op`, so the whole-word 1Password-CLI protection now fires on paths too — while `top`/`Dropbox`/`Finder` still don't over-match the `op` exception. Regression-locked via a `--denied` test hook (+6 assertions: /path/op→denied, bare op→denied, top→allowed, 1Password-path→denied).

**@157 (bind identity before sampling nice):** `cur` is now read AFTER `live_start` is bound, and `recheck_start` brackets that read up to the `renice` — a recycle can no longer pair a stale nice value with a live identity (wrong revert record / -ge decision).

Suite 37/37 (58/58 with round-3). `basename ""` edge verified safe. gitleaks clean. Responder-only.

PR #282, opened 2026-07-22
feat(bin): plugin-cache-sweeper — reclaim orphaned Claude Code plugin staging clones

ekson73:
`[PR-as-Prompt]`

## Context

The Claude Code plugin manager clones each marketplace into a staging directory under
`~/.claude/plugins/cache/` and does not remove it once the operation **succeeds**. They
accumulate without bound.

Measured on one workstation: **1318 orphaned clones (~8.0 GB) over 8 days**, ~20–30 new
per hour of active use. They were a material contributor to a full-disk condition
(0 bytes free on the data volume) that this work started from.

Reported upstream as [anthropics/claude-code#80367](https://github.com/anthropics/claude-code/issues/80367).
This script is the local stanch until that lands — and stays useful afterwards, since
crash-interrupted operations will always leave staging behind.

## Objectives

- Reclaim orphaned staging clones **safely** — five gates before any deletion.
- Be adoptable by anyone running Claude Code with plugins: one universal boundary, no
  machine-specific paths, `bash 3.2`, no `jq`.
- Encode the measurement lessons in the code, so the next reader does not re-derive them.

## Design decisions worth reviewing

**READ-ONLY by default.** Deletion requires an explicit `--apply`. A community tool that
reclaims disk should not delete on first invocation by someone who has not yet read what
the boundary is.

**Age comes from the epoch embedded in the directory NAME, not from `mtime`.** These are
two different clocks. `mtime` moves on any content change — *including a partial
deletion*. Measured: 14 orphans born over a previous week all showed an `mtime` from the
same recent minute and each held exactly 1 entry — carcasses of an `rm -rf` killed
mid-flight. An interrupted cleanup **rejuvenates its own victim**, so an `mtime` gate
permanently shields the safest garbage on disk (an emptied dir) while happily considering
intact clones. Exactly backwards. Locally this fix moved the eligible count from 0 to 49.

**One global `lsof` filtered by path, never `lsof +D <cache>`.** Measured: `+D` on a
565-clone cache exceeded 120 s and had to be killed; the global pass answering the same
question took 1 s. `+D` descends the tree and stats every file, so its cost scales with
*content* (hundreds of git clones × thousands of objects), not with candidate count.

**`freed_mb` from the free-space delta, never a `du` estimate.** On APFS a clonefile is a
separate inode sharing blocks, so `du` counts every copy in full and systematically
over-reports clone-heavy trees. In report mode `freed_mb` is pinned to `0` rather than
leaking `df` noise from other processes — a run that deleted nothing must not claim it
reclaimed space.

## The five gates

| # | Gate |
|---|---|
| 0 | basename literally starts with `temp_` |
| 1 | older than the age floor, read from the name's immutable epoch |
| 2 | a real directory, and not a symlink |
| 3 | physical path (`pwd -P`) resolves **inside** the cache boundary |
| 4 | no open file handle |

Exit is **always 0** — a housekeeping sweeper must never fail a scheduled cycle. Problems
are reported in the JSON envelope, not via exit codes.

## Acceptance

- [x] `bash -n` clean under `/bin/bash` 3.2.57 (what `launchd` actually uses)
- [x] 18/18 tests pass — including that the real installs and any symlink target outside
      the cache **survive** `--apply`, and that the fresh-`mtime` carcass is still reclaimed
- [x] Dogfooded against a real 1318-orphan cache: ~8.0 GB reclaimed across three runs,
      zero installs touched, zero orphans older than 24 h remaining
- [x] `gitleaks` clean
- [x] Scheduled hourly via `launchd` on the origin workstation and observed exiting 0 with
      empty stderr

## Out of scope (deliberately)

The two health guardians this work also produced (`disk-health-guardian`,
`system-health-guardian`) are **not** in this PR. They carry an operator-specific
DENYLIST/ALLOWLIST and would need real generalization; half-porting them here would ship
a tool whose safety boundary is tuned for someone else's machine. That is a separate,
larger piece of work.

## Cross-links

- Upstream bug: `anthropics/claude-code#80367`
- Related: `#77107` (AV interrupting plugin install mid-rename — same clone-then-rename
  area, but the interrupted path rather than the successful one)

### DIFF
--- a/bin/plugin-cache-sweeper.sh
+++ b/bin/plugin-cache-sweeper.sh
@@ -47,12 +47,18 @@ set -uo pipefail
 CACHE="${PCS_CACHE:-${CLAUDE_CONFIG_DIR:-$HOME/.claude}/plugins/cache}"
 AGE_MIN="${PCS_AGE_MIN:-180}"
 APPLY=0
+BAD_OPT=""   # set by the arg loop when an option's value is absent/empty; gate -3 refuses on it
 
 while [ $# -gt 0 ]; do
   case "$1" in
     --apply)    APPLY=1 ;;
-    --age-min)  shift; AGE_MIN="${1:-180}" ;;
-    --cache)    shift; CACHE="${1:-$CACHE}" ;;
+    # ⛔ A MISSING value must NOT fall back to the default. `--cache` with nothing after it used to
+    #    silently keep the DEFAULT cache — so `--apply --cache` (value lost to a typo, an unquoted
+    #    empty shell var, or a truncated wrapper) aimed the deletion at the operator's REAL plugin
+    #    cache while the envelope reported that path as if it had been requested. `${1:-default}`
+    #    cannot distinguish "absent" from "empty"; both are refusals here (coderabbit #282 round 2).
+    --age-min)  if [ $# -ge 2 ] && [ -n "${2:-}" ]; then shift; AGE_MIN="$1"; else BAD_OPT="--age-min"; break; fi ;;
+    --cache)    if [ $# -ge 2 ] && [ -n "${2:-}" ]; then shift; CACHE="$1";   else BAD_OPT="--cache";   break; fi ;;
     -h|--help)  sed -n '2,45p' "$0"; exit 0 ;;
     *)          printf 'unknown argument: %s (try --help)\n' "$1" >&2; exit 0 ;;
   esac
@@ -62,6 +68,14 @@ done
 # JSON-escape a string field. The cache path is caller-supplied (--cache / PCS_CACHE)
 # and a quote or backslash in it would otherwise terminate the field early and emit
 # malformed JSON to a machine consumer.
+# JSON-escape a string field: backslash, quote, tab, and embedded newlines.
+# ⛔ RAW C0 CONTROL CHARS ARE REJECTED UPSTREAM, NOT ESCAPED HERE (see the reject gate below).
+#    RFC 8259 forbids U+0000-U+001F raw inside a string, and a `\uXXXX` escaper written in awk
+#    is a TRAP: awk strings cannot hold NUL, so a 32-entry lookup table silently SHIFTS and the
+#    envelope then parses cleanly while decoding the WRONG byte (measured: \x01 decoded as \x00,
+#    \x0b as \t — a collision). Valid JSON pointing at a byte the caller never passed is worse
+#    than invalid JSON: the first fails loudly, the second lies. A cache path has no legitimate
+#    use for a control char, so the honest handling is to refuse the input (Gordian).
 json_esc() {
   printf '%s' "$1" | awk 'BEGIN{ORS=""}
     { gsub(/\\/,"\\\\"); gsub(/"/,"\\\""); gsub(/\t/,"\\t")
@@ -69,12 +83,40 @@ json_esc() {
       printf "%s", $0 }'
 }
 
+
 emit() { # emit <swept> <would> <young> <refused> <remaining> <freed_mb> <inuse> <note>
   printf '{"tool":"plugin-cache-sweeper","cache":"%s","mode":"%s","age_min":%s,' \
     "$(json_esc "$CACHE")" "$([ "$APPLY" -eq 1 ] && echo apply || echo report)" "$AGE_MIN"
   printf '"swept":%s,"would_sweep":%s,"too_young":%s,"refused":%s,' "$1" "$2" "$3" "$4"
   printf '"remaining":%s,"freed_mb":%s,"in_use":%s,"note":"%s"}\n' "$5" "$6" "$7" "$(json_esc "$8")"
 }
+
+# gate -3 — an option whose value was absent or empty. MUST come first: the later gates read
+# $CACHE/$AGE_MIN, and acting on a value the caller never supplied is the whole defect.
+if [ -n "$BAD_OPT" ]; then
+  emit 0 0 0 0 0 0 0 "missing-option-value"
+  printf 'plugin-cache-sweeper: %s requires a non-empty value — nothing swept\n' "$BAD_OPT" >&2
+  exit 0
+fi
+
+# gate -2 — refuse a CACHE path carrying a raw C0 control char. Fails CLOSED (sweeps nothing,
+# reports, exits 0) exactly like the age gate: a path we cannot faithfully serialize is a path we
+# must not act on. `tr -d` keeps the comparison byte-exact and needs no locale assumption.
+# ⛔ MUST sit AFTER emit() is defined — placed above it this called an undefined function and
+#    emitted NOTHING at all (measured: `emit: command not found`, empty stdout, so a machine
+#    consumer got no envelope instead of a refusal). A fail-closed gate that cannot report is a
+#    silent exit, not a guard.
+if [ "$(printf '%s' "$CACHE" | LC_ALL=C tr -d '\001-\037' | wc -c)" != "$(printf '%s' "$CACHE" | wc -c)" ]; then
+  # ⛔ SANITIZE BEFORE REPORTING — the refusal envelope must itself be parseable. Emitting the
+  #    offending path verbatim reproduced the exact defect being refused (measured: the gate fired
+  #    correctly, note=invalid-cache-path swept=0, yet `python json.load` still rejected the output
+  #    at the `cache` field). A guard whose own report is malformed hands the machine consumer
+  #    nothing to act on. `?` marks each removed byte so the path stays recognizable to a human.
+  CACHE="$(printf '%s' "$CACHE" | LC_ALL=C tr '\001-\037' '?')"
+  emit 0 0 0 0 0 0 0 "invalid-cache-path"
+  printf 'plugin-cache-sweeper: --cache contains a control character — nothing swept\n' >&2
+  exit 0
+fi
 
 # gate -1 — the age floor must be a non-negative integer BEFORE it reaches the
 # envelope. `age_min` is a JSON *number* field, so a non-numeric value is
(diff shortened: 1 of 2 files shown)


ekson73:
PDCA round applied in `ce783cd` — all 3 findings verified empirically **before** fixing (Mente Tomé: a bot verdict is data to measure, not inherit).

|---|---|---|---|
| 1 | `lsof` capture drops `.clone`; in-use gate fails open when lsof is unavailable/timed-out | ✅ **CONFIRMED — data-loss class** | fixed + regression |
| 2 | `--age-min` unvalidated; string fields not JSON-escaped | ✅ **CONFIRMED** | fixed + regression |
| 3 | test bootstrap: `set -euo pipefail` + source `lib/common.sh`/`lib/json-rpc.sh` | ⚠️ **PARTIALLY VALID** | `set -e` adopted; `source` declined (rationale below) |

**F1 — the sharpest catch, and it was real.** The pattern `temp_[a-z]*_[0-9]*_[a-z0-9]*` truncates `temp_subdir_<ms>_<rand>.clone`, while gate 4 compares **whole** names (`grep -qx`) — so a `.clone` candidate could never match its own INUSE entry. **Measured**: a `.clone` dir with a live file handle was **DELETED** under `--apply` (`swept:1, refused:0`). Three fixes:
- pattern admits the optional `.clone` suffix
- `grep -qx` → **`grep -Fqx`** (the `.` in `.clone` is a regex metacharacter — a non-fixed compare could match the wrong entry; this was a second latent defect your finding surfaced)
- **fail-closed `INUSE_OK` trust flag**: absent/timed-out `lsof` left `INUSE=""`, indistinguishable from "nothing is open" — the free-negative that turns a safety gate into a no-op precisely when it matters. An un-RUN probe now refuses every delete under `--apply`. ⚠️ Subtlety worth recording: lsof rc is captured **separately** from the filter, because `grep -o` exits 1 on **no match** — the *healthy* case — so piping them together would have refused everything forever. My first attempt had exactly that bug; caught before commit.

**F2 — confirmed with a crafted value.** `age_min` is a JSON *number* field, so the raw interpolation emitted malformed JSON: `"age_min":abc"; rm -rf /,`. Now gated to a non-negative integer, failing **closed** (`swept:0`, `note:"invalid-age-min"`, still `exit 0` — the never-fail-a-session guarantee is unconditional), plus `json_esc()` on `cache`/`note` (both caller-supplied via `--cache`/`PCS_CACHE`).

**F3 — half adopted, half declined with evidence.** `set -euo pipefail`: **adopted** — verified safe (26/26 pass, rc=0, identical to baseline) and it matches repo convention (5/6 tests use it). The `source lib/common.sh` + `lib/json-rpc.sh` half is **declined**: those libs live at `plugin-scripts/governance/lib/`, **not** repo-root `lib/`; only 1 of 6 tests sources them; and the sweeper is standalone-by-design (AAIF cross-vendor, POSIX bash 3.2, no jq) — sourcing a governance lib into a test whose subject does not use it would add coupling for zero verification value. The path-based instruction that produced this finding globs `**/{hooks,tests,scripts}/*.sh` and appears calibrated for `hooks/`, where those libs are genuinely the contract.
> ⚠️ **Self-correction**: my first probe of F3 was **invalid** — I copied the test *outside* `tests/`, so `HERE` resolved to a temp dir and `BIN` pointed nowhere (rc=127 = command-not-found), which I nearly misreported as "`set -e` breaks the test". Re-run in place with a positive control; `set -e` is clean.

**Shellcheck SC2015 (`A && B || C`) ×11** — acknowledged, not changed: every occurrence is the `ok`/`no` assertion idiom where `C` running on a truthy `A` is exactly the intended fallback (both branches only print + set a flag). Rewriting 11 sites to `if/then/else` would add noise without changing behavior.

**Regressions added (18 → 26 assertions):** held-open `.clone` refused **and** counted · invalid `--age-min` exits 0 / reports in envelope / stays a JSON number / sweeps nothing / leaves the candidate on disk · quoted cache path keeps the envelope parseable.
**Negative control**: reverting *only* the `.clone` capture makes the new gate-4 test **FAIL** with `swept:1` — the test earns its keep rather than passing for the wrong reason.

Verified: 26/26 ALL PASS (rc=0, captured directly — no pipe masking) · `shellcheck -S warning` clean · gitleaks clean · `validate-plugin.sh` PASSED · all 4 envelope paths parsed with `python3 json.load`.

@coderabbitai review

ekson73:
Round-2 applied in `ad0e01e` — all four findings measured by **effect**, and three were real defects. **Two of them are mine**, introduced by the round-1 fix.

| # | Finding | Verdict |
|---|---|---|
| A | `--age-min` / `--cache` fall back to the default when the value is missing | ✅ **CONFIRMED — the dangerous one** |
| B | `json_esc` must encode *every* control char, or reject | ✅ **CONFIRMED** (fixed by rejecting) |
| C | restrict `lsof` results to paths under `$BOUND` before matching | ✅ **CONFIRMED — permanent false refusals** |
| D | simulate an unavailable probe by shadowing `lsof` | ✅ accepted |

**A — a missing option VALUE aimed the deletion at the real cache.** `--cache` with nothing after it kept the **default** cache, so `--apply --cache` (value lost to a typo, an unquoted empty var, a truncated wrapper) targeted the operator's actual plugin cache while the envelope reported that path as if requested. `${1:-default}` cannot distinguish *absent* from *empty*; both are refusals now (`note:"missing-option-value"`, fail-closed, exit 0). Proven on the **effect**, not the text: value missing → an eligible orphan in the fallback cache **SURVIVED** (`swept:0`); path supplied → same fixture reclaimed (`swept:1`), the control showing the test can see a deletion at all. An assertion reading only `note` would have passed even with the files gone.

**B — my round-1 escaper left 31 holes, and my first fix was worse than the bug.** RFC 8259 forbids raw U+0000-U+001F; round-1 covered only `\t`. ⚠️ I first wrote a `\uXXXX` escaper **in awk** — which cannot hold NUL in a lookup table, so it **shifted** and emitted *valid* JSON decoding the **wrong byte** (measured: `\x01`→`\x00`, `\x0b`→`\t`). Valid JSON pointing at a byte the caller never passed fails silently; invalid JSON fails loudly. Reverted to a **refusal** — a cache path has no legitimate use for a control char.
Second self-caught defect in the same fix: the gate fired correctly yet **its own refusal envelope embedded the offending byte**, reproducing the very defect it refused. A guard whose report is malformed hands a machine consumer nothing. Now sanitized to `?` first. Also fixed placement — the gate initially sat **above** `emit()` and printed *nothing at all* (`emit: command not found`). Verified across the full C0 range: **31/31 refused with a parseable envelope**, 4/4 normal paths (quotes · backslashes · accents · spaces) **not** over-refused.

**C — the sharpest finding: the in-use set was machine-wide.** `lsof` reports every open path on the host and gate 4 compares **basenames**, so an open file in an unrelated directory that merely *shares* a candidate's name blocked the legitimate in-cache candidate **forever** (measured: namesake held outside the cache → `swept:0 refused:1 in_use:1`). Now scoped to `$BOUND/` before name-matching, keeping full paths through the boundary test and deriving names only from survivors. After: `swept:1 refused:0` — while round-1's *in-boundary* held-open `.clone` assertion still passes, proving the gate was **scoped, not loosened**. Over-refusing is safer than over-deleting, but a permanent false refusal makes the sweeper useless at its one job.

**Tests 26 → 35**, all mutation-verified. **Negative control on C**: reverting *only* the boundary scoping makes the new assertion fail with the exact predicted message (`out-of-boundary namesake blocked the in-cache candidate`); restored → 35/35 green.

`shellcheck -S warning` clean · `validate-plugin.sh` PASSED · gitleaks clean.

@coderabbitai review

ekson73:
Round-3 applied in `af5113b` — the finding **located a real defect at exactly the right line, and prescribed the wrong direction**. I fixed the defect, not the prescription.

**Requested**: require exactly 13 digits for the epoch segment in the INUSE basename filter.
**Applied**: widened the filter to `temp_*` — the same class the enumeration already admits.

**Why the direction inverts.** Gate 4 compares **whole basenames**, so the filter is not a validator — it decides *what the gate can even see*. Enumeration admits `temp_*` (`find -name 'temp_*'`, and gate 0's `case "$b" in temp_*)`), while the filter admitted only `temp_[a-z]*_[0-9]*_[a-z0-9]*(\.clone)?`. **Any candidate in the first set but not the second is invisible to the gate and gets deleted while in use.** Tightening the epoch shrinks the admitted set, so it *adds* members to that invisible set — it makes the hole bigger while reading as stricter.

**Measured on a 5-dir fixture** (each name **is** a candidate; each fails the narrow shape):

| name | why it fails the narrow shape | held open | seen by gate 4 (before) |
|---|---|---|---|
| `temp_weird` | no epoch/rand segments | ✅ | ❌ |
| `temp_UPPER_<ms>_x` | uppercase vs `[a-z]*` | ✅ | ❌ |
| `temp_a-b_<ms>_x` | hyphen vs `[a-z]*` | ✅ | ❌ |

**3 of 3 invisible.** Under `--apply`, **2 of the 3 were actually deleted with a live handle open** — and the third is the instructive one: `temp_weird` survived on the **age gate's** fail-closed branch (no parseable epoch ⇒ `too_young`), **not** on gate 4. A different gate was covering for the broken one, which is precisely why every prior test stayed green.

⚠️ **The general shape, since this is the third occurrence in this PR**: divergence between *"what we may delete"* and *"what we check for use"* is always a **silent exemption**. Round-1 was the same class one layer out (`.clone` truncated ⇒ never matched its own candidate); round-2 was its mirror (machine-wide `lsof` ⇒ permanent false *refusal*). Keep both sides on **one class** and let the later gates narrow — they **refuse**, they never **exempt**.

**Tests 35 → 36.** Negative control: reverting *only* the pattern fails the new assertion with `DELETED 2 held-open candidate(s) invisible to the in-use filter (fail-open)`; restored → **36/36**. The round-2 out-of-boundary-namesake assertion still passes, so this **widened what the gate sees without loosening the boundary**.

`shellcheck -S warning` clean · `validate-plugin.sh` PASSED · gitleaks clean.

@coderabbitai review

ekson73:
Round-4 in `81eeefd` — **confirmed and reproduced exactly as described.** Your diagnosis was correct in all three parts, including the one I would have missed: that `INUSE_OK=1` is assigned on `lsof`'s rc **before** the filter runs, so a parse failure downstream is trusted rather than caught.

Measured with `--cache '/tmp/cache|with-pipe'` and a live handle held:

```
sed: bad flag in substitute command: '\'
{"swept":1,"would_sweep":0,"refused":0,"in_use":0,"note":"ok"}   ← deleted, handle open
```

`note:"ok"` while a gate had silently collapsed — the envelope reported health.

**Fix removes the regex rather than escaping the delimiter.** Escaping `|` patches one character; a filesystem path may legally contain *any* metacharacter, so the escape set is unbounded relative to the input domain. Prefix now stripped with `${line#"$BOUND"/}` — no regex, no delimiter, nothing to escape — basename cut at the first `/`. The quotes are load-bearing: unquoted, `$BOUND` is read as a **pattern**, so `*`/`?`/`[` would strip the wrong prefix. And per your second point, any failure building the set now sets `INUSE_OK=0`, so gate 4 refuses instead of trusting a blind `""`.

⚠️ **My first cut of this fix was broken, and it looked like a pass — worth recording because the failure mode is the one this PR keeps circling.** I wrote `case` patterns without balanced parens inside `$( )`, which **bash 3.2** (this file's declared floor, and the macOS system bash) mis-parses: the substitution captured the literal rest of the source text and the script aborted on `b: unbound variable`. The held-open directory **survived** — which reads exactly like a refusal. The only thing that exposed it was the **positive control**: a plain path must still *reclaim*, and it didn't. The tool was not strict, it was **dead**.

⇒ **Survival proves refusal only when the same build can still delete.** Every new assertion is now paired with an unheld candidate that must be reclaimed in the same run; a blind control fails the test explicitly rather than passing quietly.

**Tests 36 → 42** — the regression you asked for, generalized past `|` since the root was never pipe-specific:

| cache path | held candidate | unheld control |
|---|---|---|
| `cache\|pipe` `cache*star` `cache?q` `cache[b]br` `cache sp` `cache.dot` | refused ✅ | reclaimed ✅ |

**Negative control**: restoring the `sed` form reproduces `held: DELETED / unheld: reclaimed` — the assertion fails for the right reason, and the control proves it wasn't blind. ⚠️ My *first* attempt at that control was itself invalid: the revert script threw `substring not found`, so nothing was reverted and I measured the **already-fixed** file while calling it the defect. Re-ran with the revert **verified present** before trusting the reading.

That is the fourth distinct instance on this PR of the same class — an instrument that answers a question adjacent to the one asked (`.clone` truncation → machine-wide `lsof` → narrower-than-candidate filter → now a parse-death mistaken for a gate). The invariant that keeps holding: **check that your instrument can still see a positive before believing its negative.**

`shellcheck -S warning` clean · `validate-plugin.sh` PASSED (Errors: 0) · gitleaks clean.

@coderabbitai review

PR #283, opened 2026-07-23
docs(adr-013): E6 forgetting criterion + mutation-hook DoD items

ekson73:
**[PR-as-Enhancement-Prompt]**

## Context

`ADR-013` (memory-CRUD gateway) is Accepted-design with the build pending. Its narrow API makes `archive` the **only** removal verb — *tiering-never-deletion made structural*. That is the right safety property, and it is not the whole property.

All four mutating verbs (`create` / `update` / `archive` / `index`) are **caller-initiated**. So the corpus forgets only when some agent *decides* to call `archive`. A gateway could satisfy the entire original DoD — every verb, every refusal, every read-only guarantee — while never forgetting anything on its own.

## Objectives

Add the **E6 lens** — *evaluate a memory mechanism by what it can FORGET, not only by what it can find* — and the DoD items it implies.

The gap is measured, not asserted (arXiv 2606.15903):

| Regime | Canonicalization | Intent-aware deletion |
|---|---|---|
| Deterministic | ≤5% (0% cross-lingual) | — |
| **Inscribe-time LLM** | **100%** | **0%** |
| **+ mutation-time hook** | — | **+22.6–24.1 pt** (70.1 → 93.3%) |

The paper's conclusion: *joint placement is necessary but not sufficient*. `archive`-as-only-removal-verb makes forgetting **safe**; it does not make forgetting **happen**.

## Changes

- **§"E6"** — names the gap and ties it to the `control_plane` axis of the placement router (`deterministic | inscribe-llm | mutation-hook | joint`). The router *classifies which regime an item needs*; this gateway *is the executor of that decision*. Erasable/correctable items are routed ⛔ `mutation-hook` or `joint`, **never inscribe-only** — a gateway that cannot honour that routing makes the router's hard gate unenforceable.
- **DoD item** — `update` resolves supersession *in the same operation*: archive the superseded topic + update index atomically, ledger-recorded. **Ambiguity REFUSES with a pointer** (same shape as `DEDUP_COLLISION`) rather than archiving the wrong topic — silent non-archival is unacceptable, silent wrong-archival is worse.
- **DoD test** — supersession archives **without any explicit `archive` call**; N superseded facts converge to N archived + 1 live *through normal writes alone*. A build passing every other test and failing this one scores the paper's 0%.

## Acceptance

- [x] DRY-checked — zero prior mentions of `E6` / `mutation-hook` / `intent-aware` / `forget` in ADR-013
- [x] Status unchanged (Accepted — design; build still pending); no API/verb change, no runtime effect until the build
- [x] Additive only — no existing DoD item weakened or removed
- [x] gitleaks clean
- [x] Amendment recorded in §Provenance with empirical basis + companion decisions
- [ ] Bot convergence

## Cross-links

- Companion (akasha): `docs/decisions/ADR-memory-placement-router-2026-07-22.md` — the router that classifies `control_plane` per item · PR `ekson73/akasha-claude#276`
- Hub: `docs/decisions/memory-architecture-roadmap-2026-07-15.md` §4 (5 components) / §6 Phase-3b
- Inherited bounds: ADR-012 (sweep — the *detector*) · Mnemosyne (the *judge*) · this gateway (the *executor*)
- Tracked: EKO-116

---
> 🤖 agent-created — `Claude Opus 4.8`, session `318de395`, under the operator's standing continue-directive. Tracker `creator` = MCP-auth account (operator), **not** the author. Unsigned commit per operator `[C07]` HITL auth 2026-06-24 (1P SSH-signing agent down).


ekson73:
@coderabbitai — all 4 findings actioned in `b67f3f1`. One of them I resolved *differently* than suggested, with evidence; details below.

| # | Finding | Action |
|---|---|---|
| L82 | Metric framing wrong | **Fixed — but to a third value.** See below. |
| L86-92 | `update` supersession identification undefined | Fixed — explicit rule added |
| L92-93 | Crash-safe atomicity undefined | Fixed — new DoD item with write ordering + recovery |
| L99 | Provenance lacks ISO 8601 timestamp | Fixed — `2026-07-23T12:32:18Z` |

---

### L82 — I verified the source rather than adopting either reading

You were right that the framing was wrong. But your proposed correction ("the 70.1 → 93.3% figures are the non-primitive-existence subset") is **also not what the abstract says**. I fetched arXiv 2606.15903 directly before editing. What it actually reports:

| Regime | Figure (verbatim from abstract) |
|---|---|
| deterministic | "5% on identifier-obfuscation, 0% on cross-lingual" |
| inscribe-time LLM | "recovers canonicalization (100%)" … "cannot help intent-aware deletion (0% on prefix-collision and compound-fact)" |
| mutation-time hook | "recovers intent-aware deletion (78-85%)" … "91.7-93.2% overall" |
| joint-placement replication | "+27.8 pt" |

**`70.1`, `93.3` and `+22.6-24.1 pt` appear nowhere in the abstract.** They were mine, not yours — an unverified figure carried in from the tracking ticket. Withdrawn, not silently swapped: the Provenance section now records the withdrawal explicitly, because a governance ADR that quietly changes a cited number is worse than one that shows the correction.

The corrected figures happen to make the argument *stronger*: the E6 thesis is "caller-initiated verbs sit in the inscribe-time regime", and **0% → 78-85% intent-aware deletion** is exactly that delta, stated directly by the source. I also tightened "≤5% canonicalization" → "5% on identifier-obfuscation", which was imprecise in the same way.

I did not adopt the subset characterization because I could not verify it from the abstract. If it comes from a table in the full text, point me at the section and I will re-check.

### L86-92 — supersession identification

The DoD said "ambiguous supersession refuses" without ever saying **how the superseded topic is found**, which made the refusal unenforceable. Now:

- `supersedes: <topic-id>` passed by the caller → authoritative, archive exactly that
- omitted → match on the **declared identity key** (`name` + `node_type` — the same key `create` already dedups on)
- exactly-one match → deterministic archive + index + ledger, same operation
- zero matches → normal `create` path
- ≥2 candidates, or fuzzy/semantic-only → **refuse with candidate ids** (`DEDUP_COLLISION` shape) so the caller re-issues explicitly

### L92-93 — crash-safe atomicity

Filesystem writes are not transactional, so "atomically" was a promise with no mechanism. Now an ordered commit with a marker:

`corpus → ledger-intent → index → ledger-commit`

That leaves exactly two recoverable states — *intent-without-commit* (startup replays the index rebuild idempotently, then closes it) or *no-intent* (nothing observable happened). **Startup MUST replay unclosed intents before serving any verb**: a gateway that serves reads over a half-applied supersession reports a corpus state that never existed. Failure injection at each of the four stages is in the DoD.

Could you please re-review?

ekson73:
@coderabbitai review

Fixes for all 4 findings landed in `b67f3f1b` (current PR head) ~15 min ago; no re-review has fired since. Requesting an explicit pass over the fix commit.

Note on the L82 finding: I corrected it to the arXiv abstract's own figures (78-85% intent-aware deletion / 91.7-93.2% overall) rather than to the subset framing you proposed, because neither my original numbers nor that characterization appear in the abstract. Rationale and verbatim quotes are in the comment above — if the subset reading comes from a table in the full text, point me at the section and I'll re-check.

ekson73:
@coderabbitai — you were right on both, and on the first one I was wrong twice. Fixed in `90583d7`.

### 1. Crash-recovery ordering — corrected to intent-first

You're right: the durable intent must precede the corpus mutation. My order (`corpus-move → intent → index → commit`) left an unrecoverable window — a crash after the archive move but before the intent write produces an observable archived topic with no intent to replay, silently contradicting the "no-intent = nothing observable" guarantee. Now write-ahead ordered:

`ledger-intent (unclosed) → corpus-move → index → ledger-commit`

Crash states reduce to exactly two, both recoverable: *no-intent* (crash before step 1 — nothing mutated) or *intent-without-commit* (the durable intent names what to finish; startup re-applies archive+index idempotently, then closes it). Failure injection now explicitly covers the corpus-move/index interval.

### 2. The metric withdrawal — I reverse it; you were right in round 1

This one I owe you a clean acknowledgement. I withdrew `70.1→93.3% / +22.6–24.1 pt` as "appearing nowhere in the source." **That was false**, and I've restored the figures.

What actually happened, because the failure is instructive:

- My round-1 verification fetched `arxiv.org/abs/` and the summary **silently dropped the subset clause** — it returned `78–85%` and `91.7–93.2%` but not the `70.1→93.3` sentence. I read that silence as absence.
- My round-1 "fix" then confidently withdrew the correct numbers.
- My attempted re-verification fetched `arxiv.org/pdf/` — which **false-negatived even on `78–85` and `91.7–93.2`**, strings I *knew* were in the abstract. That should have told me the probe was broken; instead it nearly confirmed my error.
- Only a third fetch (ar5iv), **gated on a positive control** (it had to find the known-present `91.7`/`78` before I trusted any of its negatives), surfaced the truth.

The abstract + Table 2 do carry both metric sets, exactly as you said:

> *abstract:* "a capability lift of ++22.6 to ++24.1 pt on the 345 non-primitive-existence cases (deterministic 70.1–70.7% → hook 93.3–94.2%)"
> *Table 2:* Lethe+LLM 322/345 (93.3%), LangGraph+LLM 325/345 (94.2%); footnote "we report the n=345 subset as the headline."

So the headline is the 345-subset lift (70.1–70.7 → 93.3–94.2, +22.6–24.1 pt); 78–85% is intent-aware deletion specifically; 91.7–93.2% is overall incl. compound_fact. All three are now in the ADR, correctly labeled, and the provenance carries a verification-error log so the next agent doesn't repeat the sub-reaching-probe mistake.

Your round-1 framing (subset vs overall, keep them distinct) was correct and I dismissed it on an incomplete probe. Thanks for holding the line through the re-review.

Could you please re-review `90583d7`?

ekson73:
@coderabbitai Both actionable comments on `90583d7` addressed in `2aa6592` — verified against the live file, both were correct.

| Finding | Fix | Commit |
|---|---|---|
| Line 92 — identity/supersession contract used inconsistent `title/slug` (create dedup) vs `name + node_type` (supersession match); `supersedes: <topic-id>` undefined vs slug-based API | Unified into ONE canonical key `(type, normalize(name))` used identically by create-dedup and implicit-supersession match; **slug = that key's addressable projection**; `supersedes: <slug>` now maps explicitly onto the slug-addressed verbs (`read`/`update`/`archive`) | `2aa6592` |
| Line 93 — E6 write-ahead path preserved intent-first ordering + startup replay but omitted durability | Added the barriers: `fsync` file **+** `fsync` containing directory (rename/create durability) **+** atomic write-temp-then-`rename` for the index; each stage reaches stable storage before the next. Rationale added inline: ordering-in-logic ≠ ordering-on-disk, so without barriers write-reordering reintroduces the archived-topic-with-no-intent unrecoverable state | `2aa6592` |

`validate-plugin.sh`: PASSED (0 errors). The `code/snyk` check failure is "Code test limit reached" (account-level quota, not a code finding — non-required/UNSTABLE).

Could you please re-review `2aa6592`?

ekson73:
@coderabbitai review

Round-4 fixes pushed in `98d0854` — all 5 actionable findings addressed + the operator-ratified `purge` policy:

| Finding | Fix |
|---|---|
| Line 27/92 — type-scoped identity not addressable | Canonical slug is now **`<type>/<normalize(name)>`** (create-arg + `--type` form it); `(user,profile)` → `user/profile` vs `(project,profile)` → `project/profile` no longer collide. Removed the buggy "slug collision **is** identity collision" assertion. |
| Line 92 — `update` request shape | Positional `<slug>` = the existing topic patched **in place**; *supersession* is the separate mode (keyed by `supersedes:`/`(type,name)`, never the positional slug); zero-match ⇒ normal `create` path. |
| Line 93 — new-live-topic durable stage | New-fact write is now an explicit durable, idempotently-replayable stage (4-step → **5-step** WAL ordering). |
| Line 93 — torn-record & rename-dir recovery | Archive `rename` fsyncs **both** source+dest directories; startup replay **discards a trailing partial JSONL record** before replaying intents. |
| Line 94 — assert superseded absent from reads | E6 test now asserts every superseded slug is **absent from active reads/search** (not just N archived + 1 live); archived copy stays retrievable. |

Also: pinned **`arXiv:2606.15903v2` (2026-06-16)** — resolving your MCP "cite the exact source/appendix" note (v2 abstract + Table 2 carry the 345 non-primitive-existence headline figures); and recorded the operator's ratified `purge` scope-boundary (archive-only corpus; `purge` = dormant, PII-only, out of the narrow API / not in MVP).

`validate-plugin.sh`: PASSED (0 errors). Could you please re-review?

ekson73:
@coderabbitai — all 3 actionable Major findings addressed in `3018dc8`.

| # | Finding (@98d0854) | Fix |
|---|---|---|
| 1 | Implicit supersession matched "the corpus" → could hit an **archived** copy → no-op archive + un-superseded live topic | Implicit match now scoped to the **active (non-archived) corpus only**; archived targets require an explicit `supersedes` reference |
| 2 | Durable intent stored only identity + id-pair; new-topic body arrives via `stdin` → crash before stage 2 leaves replay unable to reconstruct the fact | Intent now persists the **full new-topic payload** (or a staged-file ref + integrity hash), fsync'd before intent-durable; **startup replay reconstructs from the intent, never from `stdin`** |
| 3 | WAL is crash-safe but **not concurrency-safe** — two concurrent writers can race the same live topic | The whole supersession transaction (implicit-match → 5 WAL stages → commit → recovery) now runs under an **exclusive corpus/ledger lock or CAS guard**; refuse-on-contention = the concurrent-writer refusal the test suite already asserts |

All three sharpen the E6 forgetting-criterion contract (the EKO-116 build target). `validate-plugin.sh` PASSED (0 errors). Could you please re-review?

ekson73:
@coderabbitai — both actionable Major findings addressed in `c3f982f`.

| # | Finding (@3018dc8) | Fix |
|---|---|---|
| 1 | `create <topic-slug> [--type …]` left `--type` optional, but the canonical key `(type,name)` and the type-qualified slug both require it → ambiguous key/slug for a typeless `create` | **`--type` is now REQUIRED (no default)** — `type` is semantic and both the key and the slug are undefined without it; a `create` missing `--type` refuses |
| 2 | WAL step 2 ("write the new live topic, its own durable stage") wasn't explicitly atomic — a plain in-place write can be torn by a crash, unlike the index write which is atomic | **New-topic write is now explicit atomic write-temp-then-`rename`** (write temp → `fsync` → `rename` into live path → `fsync` dir), matching the index write's atomicity |

Both complete the E6 crash-safety contract's own logic (the EKO-116 build target). `validate-plugin.sh` PASSED (0 errors). Could you please re-review?

PR #284, opened 2026-07-23
feat(EKO-116): memory-CRUD gateway (ADR-013 build phase)

ekson73:
## [PR-as-Documentation-Prompt] feat(EKO-116): memory-CRUD gateway (ADR-013 build phase)

Implements the ADR-013 **memory-CRUD gateway** — one canonical owner of memory mutations; every other actor must ask it. POSIX bash 3.2 + jq, no runtime deps beyond `jq`.

### What this adds

| Artifact | Lines | Role |
|---|---|---|
| `bin/memory-gateway` | 861 | 7-verb narrow API (create/update/archive/index + read/search/neighborhood) |
| `tests/test-memory-gateway.sh` | 635 | TDD suite — 42 test blocks / 57 assertions (every verb, dedup, concurrency, boundary, archive-not-delete, index regen, read-only, E6, WAL crash-safety, exit-codes, audit ledger + red-team + qodo/CodeRabbit regression pins) |
| `skills/memory-gateway/SKILL.md` | 112 | agent-facing wrapper (contract + examples) |

### The 7-verb narrow API
- **Mutating** (locked, WAL-atomic): `create` · `update` · `archive` · `index`
- **Read-only** (no lock, never mutate): `read` · `search` · `neighborhood`
- **Canonical identity** `(type, normalize(name))`; slug = `<type>/<normalize(name)>` (case-fold + kebab). `(user, profile)` → `user/profile` and `(project, profile)` → `project/profile` are distinct addresses that never collide — used identically by `create` dedup and by implicit E6 supersession match.

### E6 mutation-hook (supersession in the same op)
`update --supersede` archives the superseded topic in the same operation — explicit `--supersedes <slug>` (authoritative) OR implicit `(type,name)` match **against the active corpus only** (archived copies excluded from candidate selection). `0` matches → normal create; `≥2` or fuzzy → refuse-with-pointer (`SUPERSEDE_AMBIGUOUS`, same shape as `DEDUP_COLLISION`). Every superseded slug is absent from active `read`/`search`/`neighborhood` while remaining retrievable from the archive tier.

### Crash-safety (5-stage WAL)
`intent (fsync'd first, carries payload + sha256, replay source of record — never stdin) → archive-superseded → write-new-topic (temp→rename+fsync, dir fsync'd) → rebuild-index (atomic temp→rename) → commit-marker`. Startup recovery replays unclosed intents idempotently and discards a trailing partial ledger line. Entire supersession transaction runs under an exclusive mkdir-mutex corpus lock (dead-pid reclaim; a writer that can't acquire within timeout **refuses** rather than racing).

### I/O contract [C06]
stdout = **exactly ONE JSON envelope** `{status, verb, slug, data, reason}` per execution · stderr = human-only · exit `0`=ok / `1`=refused-or-usage / `2`=setup. Enforced by regression pins (see below).

---

## Verification (build phase)

- `bash -n` clean on both `bin/memory-gateway` and the suite.
- **57/57 assertions deterministic across 23 consecutive runs** (incl. runs with an intentionally-open inherited stdin), including WAL failure-injection at each of the 5 stages + the new EPERM/ESRCH liveness-disambiguation and jq-fallback pins. A pre-existing test-harness transient (a swallowed fork-EAGAIN in the `jqr` jq-helper under burst load — the gateway envelope was always correct) was root-caused and fixed with a bounded retry-on-empty (provably non-masking: `jq -r` prints `null` not `""` for null/absent, and no assertion compares to literal `""`).
- `tests/validate-plugin.sh`: PASSED (0 errors; 1 pre-existing unrelated warning).
- `gitleaks detect --no-git --source .`: no leaks found. Fixtures are synthetic only.

## Adversarial review trail (Elenchus mandatory red-team — H3 irreversible∧blast)

This change is HIGH criticality (a canonical memory mutator), so it went through independent adversarial red-team rounds (verifier > generator; each fix re-verified). The final round ran the **raw binary** (not the suite) on isolated `mktemp -d` corpora with an isolated audit ledger, independently reproduced #1-closed (fifo-wedge: the blocked writer holds no lock, a concurrent writer proceeds rc=0) + #3-closed (single envelope on `index`) + supersede-recovers-through-the-drain, and returned **CLEAR** (no data-integrity / crash-safety / availability / [C06] regression on the changed surface). Every finding is recorded here — fixed, accepted, or deferred — none silently dropped.

### Fixed
| # | Severity | Finding | Root-fix |
|---|---|---|---|
| pass-1 | CRITICAL | validation ran inside a `$()` subshell → a boundary `_refuse` (exit 1) could only kill the subshell, letting a **second JSON envelope** print at exit 0 (violates the [C06] one-envelope contract) | Refactored `_atomic_write` to an **arg-based** call that runs in the caller's shell, so a boundary refuse aborts the whole process. Pin 24 (exit-codes/envelope shape) + pin 36 guard it. |
| pass-2 #1 | MED | the blocking stdin `$(cat)` sat **inside** the single-writer lock → an open-but-silent stdin could wedge the lock (availability) | Drain the body **off-lock in dispatch** before `_lock_acquire` (`create\|update` only; archive/index take no body). Validation still refuses-fast off-lock (keeps the test-29 valueless-flag DoS guard). Mutation semantics under the lock are byte-identical. Pin 35 (wedge repro) guards it. |
| pass-2 #3 | LOW | `do_index` read the rebuild count via `$()`, re-nesting `_atomic_write` in a command-substitution subshell (same double-envelope class as pass-1) | `_rebuild_index` exports a `REBUILD_COUNT` global; `do_index` reads it in-shell (no `$()`), so `_atomic_write`'s boundary refuse aborts the process. Pin 36 (single-envelope on `index`) guards it. |
| test-harness | (flake, not the binary) | test 22 helper fed the body via a `BODY=…` env var with no stdin pipe → the gateway read the suite's **inherited** stdin and intermittently hung | Pipe the body via stdin (same class already fixed for test 34). Reproduced under a deliberately-open inherited stdin → 3/3 pass. |

### Accepted (documented — not a defect against the spec)
- **pass-2 #2 — trailing-newline normalization on the stored body.** ADR-013's I/O contract specifies the JSON-envelope shape and exit codes but has **no byte-exact / round-trip / trailing-whitespace fidelity requirement** on the stored topic body. Normalizing a trailing newline is within spec; accepted as-is.
- **pass-3 behavioral note — `create`/`update` with an *open-but-silent* stdin self-stalls before an early refuse.** Because the off-lock body drain (the #1 fix) precedes the verb's arg-validation, an EOF/piped stdin — **the agent case** — still refuses bad args (USAGE/NOT_FOUND/DEDUP/SUPERSEDE_AMBIGUOUS) fast, but a mis-wired caller that leaves stdin open-and-silent self-stalls on the drain until EOF. This is the deliberate, benign trade for the no-wedge property: the stall is **off-lock and self-only** (it never touches the corpus lock, never wedges another writer, never affects the corpus) — a strict improvement over the pre-#1 state where the same stdin wedged *every* writer with `CONCURRENT_WRITER`. Only bites human interactive debugging (agents pipe+EOF). Documented in the dispatch-drain comment.

### Deferred (tracked, low-impact — not blocking the build DoD)
- **#8b — audit-ledger lock steal.** The audit-append path does not itself steal a stale lock; only the corpus mutex does. Ledger contention is bounded by the corpus mutex that already serializes mutators, so this is a theoretical residual.
- **#8c — audit-ledger fsync.** The ledger append is not independently fsync'd (the *intent* record is, before any mutation — that is the replay source of record). A crash can lose only the trailing *audit* line, which recovery already discards if partial; no corpus state is affected.
- **#8d — jq-missing → no JSON envelope.** If `jq` is absent, setup fails at exit 2 with a stderr diagnostic rather than a JSON envelope (jq is the envelope emitter). Documented as a setup-tier (exit 2) condition.
- **#4 notes** — `_replay` occupancy edge, the µs-window between lock-check and acquire (closed by the mkdir-mutex atomicity), and NUL-in-body handling — all negligible / already-covered; recorded for completeness.
- **pass-3 optional — stdin read-timeout on the drain to restore fast-fail on open-silent stdin.** Considered and deferred: a *naive total-time* timeout would risk truncating a legitimately-slow large body (silent data-loss — strictly worse than the benign self-stall it removes); a correct fix needs an *idle-only* timeout that distinguishes "no bytes ever" from "slow producer." Out of scope for the build DoD; only affects human interactive debugging (agents pipe+EOF). The red-team explicitly did not gate the merge on it.

---

## PDCA cycle — qodo + CodeRabbit review

After the Elenchus red-team CLEARED my own diffs, **qodo + CodeRabbit reviewed the whole gateway surface** (multi-lens C3 convergence — they examine the pre-existing surface the red-team's changed-scope did not). 11 findings surfaced; each was verified against the real code (Mente Tomé — bots false-positive too), then **fixed** with a minimal root-cause diff + a regression pin, or **deferred with rationale** (none silently dropped). Fixes committed in `850bbbc`; 6 new regression tests (37–42) pin them (57 assertions total, 20/20 deterministic).

### Fixed (8)
| # | Bot | Finding | Root-fix + pin |
|---|---|---|---|
| [122] | qodo | **Reclaim lock could delete the owner** — a stale-lock reclaim `rm`'d the lock dir, but a peer could reacquire between the stale-check and the `rm` (TOCTOU) → the peer's live lock was destroyed | `_lock_acquire` steals via **atomic `mv "$LOCK" "$LOCK.stale.$$"`** (renames the dir it observed; never `rm`s a dir a peer may now own) + `_audit` acquired-flag. Pin: test 37. |
| [154] | qodo | **Read ignored the boundary symlink** — `read`/`search`/`neighborhood` did not re-check the corpus boundary, so a symlinked slug could resolve outside the corpus root | `_assert_read_safe` (realpath-within-root **and** not-a-symlink) guards all 3 read-only verbs. Pin: test 39. |
| [189] | qodo | **I/O failures masked as `ok`** — a failed `mkdir`/`mktemp`/`printf`/`mv` inside `_atomic_write` could still print a `status:ok` envelope | `_err_io` emits **exit-2 `IO_ERROR`** (never a false `ok`); every `_atomic_write`/`_archive_topic` mutation checks `\|\| return 1` / `\|\| _err_io`. The still-open WAL intent is completed idempotently by the next invocation's sha256-guarded recovery. Pin: test 39 (I/O-fail → exit 2, nothing persisted). |
| [225] | qodo | **Frontmatter accepts newline injection** — a `\n`/`\r` in `--name`/`--description` could inject extra frontmatter lines | `_no_ctrl_lines` rejects control chars with `USAGE`. Pin: test 40. |
| [64] | qodo | **E6 test misses `neighborhood`** (also a real code gap) — the headline forgetting test never asserted a superseded slug is absent from `neighborhood` | **Code + test**: `do_neighborhood` gained an E6 liveness gate (`[ -f "$CORPUS/$l.md" ] \|\| continue`) so a superseded/archived slug is never returned as a neighbor; new test 38 asserts it. |
| [258] | CodeRabbit | **EPERM ≠ ESRCH** — `kill -0` returns non-zero for a dead pid **and** for a live pid owned by another UID → a live holder could be judged stale and evicted | `_lock_is_stale` disambiguates with a **`ps -p` probe**: only a pid `ps` **also** can't see is truly dead (fast reclaim); a live-other-UID holder (or `ps` unavailable) falls through to the ts-age backstop — **fail-toward-safe** (never evict an unproven-dead holder → no double-writer). Pin: test 41. |
| [93] | qodo | **jq path lacks fallbacks** — a minimal PATH could fail to resolve `jq`, leaving no envelope | `_resolve_bin` (`command -v` → absolute candidates incl. mise/asdf shims) resolves `jq`/`python3`/`ps` robustly. Pin: test 42 (absolute fallback resolves + runs under empty PATH). |
| [381] | CodeRabbit | **Unrestricted direct edits in SKILL.md** — the Advisory Posture had an over-broad "does not hard-block direct edits" exception | Tightened: identity-change/supersede are **gateway-only**; a direct edit is tolerated only as a last-resort **quiesced** migration + `index` reindex + re-read verification. |

### Deferred (3, with rationale — tracked, not blocking the build DoD)
- **[3] qodo — `purge` verb not implemented.** ADR-013 explicit non-goal: forgetting = archive-tier supersession (E6), always retrievable via `.archive/`. A compliant hard-erasure/purge (authorized, audited, non-automatic) is a distinct future ADR item, not this gateway's scope.
- **[33] qodo — `_topic_content` lacks bi-temporal frontmatter (`valid_until`/`superseded_by`).** ADR-013 chose archive-tier + WAL supersession over bi-temporal metadata. E6-forgetting **is** enforced — by the archive move + the read/search/neighborhood liveness gates — just not via frontmatter fields. A design alternative, not a defect in the chosen design.
- **[349] CodeRabbit — WAL never compacted.** CodeRabbit itself: *"Correctness is unaffected"* (trivial/nitpick). Compaction must coordinate with the corpus lock to avoid racing a concurrent appender — mis-done, it risks the very durability the WAL provides. Steady-state-only cost on a long-lived corpus → premature optimization (Knuth/measure-first). Deferred with a tracked note.

### PDCA pass-5 — CodeRabbit re-review on `850bbbc`
CodeRabbit re-reviewed the new head and returned **CHANGES_REQUESTED** (not a stale review). 1 Major + 1 Minor, both **genuine** (verified against real code — Mente Tomé), fixed in `4af469a`; the WAL nitpick re-surfaced as 🔵 Trivial and was re-deferred.

| # | Bot | Finding | Root-fix + pin |
|---|---|---|---|
| [763] | CodeRabbit | **[Major] neighborhood leaks a symlink-escaping neighbor slug** — `do_neighborhood` recorded a discovered neighbor after only `[ -f "$CORPUS/$l.md" ]`, which *follows* symlinks. A `_valid_slug` (string-safe) neighbor whose type-dir or topic file is a symlink escaping the corpus was surfaced. A **leaf** neighbor is listed but never promoted to a `$node`, so the line-763 node-guard could never catch it. | Apply the **same already-verified `_assert_read_safe`** at neighbor-discovery → read/search/neighborhood now share ONE realpath boundary over every surfaced item (also stops the escaping slug entering the frontier). No content ever leaked (neighborhood returns slug strings only); this closes a boundary-consistency gap + a weak existence-oracle. Pin: **test 43** (proven non-tautological — FAILS without the guard, escaped slug leaks; PASSES with it + a legit positive control). |
| [606/626] | CodeRabbit | **[Minor·SC2015] `A && B \|\| C` isn't if/else** — in tests 40/41, `no` could double-fire if `ok` returned non-zero. | Converted both assertions to explicit `if/then/else`. |

**Verification:** 58/58 assertions PASS; test 43 catches the bug when the guard is reverted (overlay-and-restore proof, SHA-verified restore). **No fresh Elenchus round** — the fix *adds* an already-red-teamed control at one more call site (no new attack surface); economic-stop (≤3–4 rounds) + multi-lens convergence (CodeRabbit + Tomé + regression pin) is the independent verification.

> The SkillSpector **MP2 "context-window stuffing"** sub-warning on the SKILL.md edit is a **false-positive** — the edit *removed* an over-broad exception and tightened wording (net +2 lines), it does not inflate context.

---

## Definition of Done (ADR-013 build phase)
- [x] `bin/memory-gateway` — 7-verb narrow API (bash 3.2 + jq; [C06] envelope + exit codes)
- [x] Test suite — every verb, dedup-refusal, concurrent-writer refusal, boundary refusal, archive-not-delete, index regeneration, read-only guarantee
- [x] E6 mutation-hook — supersession in the same op (explicit + implicit-active-only; ambiguous/fuzzy refuse-with-pointer)
- [x] E6 atomicity — crash-safe across topic/index/ledger (intent-first WAL; failure-injected at each of the 5 stages)
- [x] E6 headline forgetting test — N superseded → N archived + 1 live through normal writes; superseded slugs absent from active reads
- [x] `skills/memory-gateway/SKILL.md` wrapper
- [ ] ≥2 ratified dogfood cycles (`dogfood-mark`) — **next, on merge** (post-merge, before any Phase-B hook)

Closes EKO-116 (build phase) once the ≥2 dogfood cycles land.

## Refs
ADR-013 (`docs/adrs/ADR-013-memory-crud-gateway.md`) · EKO-116

### PDCA pass-6 — qodo re-review on `4af469a` (I/O-mask finding [189] not fully closed)

qodo correctly caught that finding [189] (I/O failures masked) was **not fully closed**: the archive step of a supersession ran inside a command-substitution subshell, so `_err_io`'s `exit 2` was swallowed — a failed archive-`mv` let the parent overwrite the un-archived original + emit a false `ok`. Verified genuine against the real code before fixing.

| Finding | Verdict | Root-cause fix | Commit |
|---|---|---|---|
| [189] I/O mask — archive path via `$(_archive_if_stale …)` swallows `_err_io` exit 2 | ✅ **genuine** (r=0 status=ok, original clobbered on pre-fix) | `_archive_topic`/`_archive_if_stale` now set a global `$ARCHSEQ` instead of `printf`-via-`$()`, so they run in the caller's process and `exit 2` propagates. Fixes **all 3** affected sites: `_wal_supersede` STAGE 2 (:471, flagged), `do_archive` verb (:693, same latent bug qodo's line-471 view didn't see), and `_replay_supersede` recovery (:425, whose old `>/dev/null` even ate the `[C06]` error envelope). | `72477ad` |
| bookkeeping: pass-4 reply-table pin numbers stale | ✅ corrected below | — | — |

**Non-tautology proof (gold-standard):** new regression pin **test 44** injects an archive-move `EACCES` during a supersession (target `.archive` dir `chmod 555`) and asserts exit-2 `IO_ERROR` + the original still live (successor never overwrote it). Verified it **fails on the faithful pre-fix code** (`git HEAD:bin/memory-gateway`) with **only test 44 red** (`r44=0 s44=ok orig_kept=0`) and passes on the fix — zero collateral. It complements test 39, which covered only the direct topic-**write** failure, not the archive-**move** path.

**Corrected pin numbers** (qodo's bookkeeping note — the pass-4 reply table cited stale numbers): symlink-escape write-boundary = **test 31** · newline injection = **test 37** · stale-lock reclaim = **tests 40/41** · read-side symlink-neighbor exclusion = **test 43** · archive-move I/O failure = **test 44**.

Tests: **59/59** passing. gitleaks clean.

### PDCA pass-7 — CodeRabbit re-review on `72477ad` → fix `61439a4`

CodeRabbit's COMMENTED review of `72477ad` surfaced one outside-diff **Major** finding on `_archive_topic` (`bin/memory-gateway:363-375`): a symlinked `.archive` could send `mv` outside `CORPUS`, and the bare `mkdir -p` lacked an `_err_io` fail-fast. **Verified genuine** (Tomé) — `_atomic_write` guards its write path with `_assert_within_corpus` (test 31), but the archive-move path did not. Fixed in `61439a4` mirroring `_atomic_write` exactly: `mkdir -p … || _err_io` + `_assert_within_corpus "$dst" || _refuse … BOUNDARY` before the `mv`. Regression pin **test 45** (non-tautology proven: only 45 red on pre-fix code, zero collateral). **No separate red-team round** (Elenchus economic-stop — adds an already-verified boundary control at one more site, not new attack surface). **60 assertions passing**, gitleaks clean.

### PDCA pass-8 — CodeRabbit re-review on `61439a4` → refine `e9beb98` + qodo whole-file triage

**CodeRabbit (1 actionable, fixed).** Re-review of `61439a4` asked to validate the archive path **before any filesystem change** (the PDCA-7 check ran *after* `mkdir`, leaving an empty escaped dir before refusing). Applied in `e9beb98`: new helper `_assert_new_path_within_corpus` resolves the **deepest existing ancestor** of the not-yet-created `$adir` and refuses a symlinked `.archive` **before** `mkdir` — zero fs artifact. Test 45 strengthened: runs **as root too** (symlink boundary is UID-independent via `cd`+`pwd -P`, unlike test 44's `chmod 555` perm-bypass) and asserts the outside dir has **ZERO descendants** (`find -mindepth 1` empty), not just no `*.md`. Non-tautology proven: overlaying the pre-refinement (`61439a4`) bin turns **only** `noart45` red (`r45=1 c45=BOUNDARY kept45=1`), 59 others green.

> **Deferred follow-up (transparency):** `_archive_topic` is now *stricter* than `_atomic_write` (still post-mkdir). Harmonizing `_atomic_write` to the same before-mkdir invariant is a sound consistency follow-up but is **kept out of this PR** to avoid touching the already-red-teamed write path (tests 31/44). Noted, not silently dropped.

**qodo (8 findings, whole-file re-review on `4d64db6`, COMMENTED/advisory) — triaged, verified against the real code (Mente Tomé):**

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | `purge` not implemented | out-of-scope **by design** | ADR-013 l.37 (tiering-never-deletion; MVP no prod PII) |
| 2 | missing bi-temporal fields | **different ratified design** | ADR-013 §E6 = archive-based forgetting, not `superseded_by` live-pointer |
| 3 | E6 test misses neighborhood | already tested | **test 38** |
| 4 | jq path lacks fallbacks | already implemented | `_resolve_bin` + **test 42** (named `qodo#4`, empty-PATH) |
| 5 | reclaim lock apaga o dono (Bug) | **false positive** | atomic `mv` rename-steal; `rm -rf` on PID-quarantine only; **tests 40/41** |
| 6 | read ignores boundary symlink (Bug/Sec) | already enforced | `_assert_read_safe` at l.582/594/779/785; **test 43** |
| 7 | I/O failures masked (Bug) | already handled | `\|\| _err_io` ×5 + `$ARCHSEQ` no-subshell; **tests 39/44** |
| 8 | frontmatter newline injection (Bug) | already refused | `_no_ctrl_lines` l.544/545/646; **test 37** |

Net: #1–#2 are ADR-013-ratified scope; #3–#8 pinned by named tests from prior PDCA rounds — no code change required. **60/60** passing, gitleaks clean, HEAD `e9beb98`.

**pass-8 addendum — CodeRabbit re-review of `e9beb98` confirmed the code fix + threat-model doc (`474bd19`).** CodeRabbit's re-review of `e9beb98` verified the before-mkdir archive fix is correct (BOUNDARY refusal before any fs change, test 45 as-root + descendant-free, deferred `_atomic_write` asymmetry accurately noted). Its one residual point was a **TOCTOU** observation — a *non-cooperating* process with write access to `$CORPUS` could swap a validated ancestor between the check and `mkdir`/`mv`. That race is **out of scope by design and unclosable at the app layer** (such a process can `rm -rf $CORPUS` outright); the real control is OS-level filesystem ownership. Per CodeRabbit's stated close-condition, `474bd19` makes the invariant **explicit** (doc-only, no logic change): a `BOUNDARY THREAT MODEL` comment block above the three boundary helpers in `bin/memory-gateway` + the `Boundary-safe` guarantee in `SKILL.md` — in-scope = pre-planted symlink under the ADR-013 single-writer contract; out-of-scope = concurrent non-cooperating FS mutator, defended only by the deploy invariant that `$CORPUS` is owned/writable by the gateway uid alone. **60/60** passing, gitleaks clean, HEAD `474bd19`.

### DIFF
--- a/bin/memory-gateway
+++ b/bin/memory-gateway
@@ -215,6 +215,19 @@ PY
   fi
 }
 
+# BOUNDARY THREAT MODEL (the 3 helpers below).
+#   These realpath guards defend the DELIBERATE-scope threat: a *pre-planted* symlink already in the
+#   corpus tree (a symlinked type-dir/topic-file/.archive) that would send a write/read/mv OUTSIDE
+#   CORPUS. That defense is exact under this gateway's core ADR-013 contract: the gateway is the ONE
+#   writer to the corpus, so the corpus tree is not being mutated by a cooperating peer mid-op (the
+#   corpus lock serializes gateway instances).
+#   OUT of scope — and NOT closable at this layer: a TOCTOU race by a *non-cooperating* process that
+#   holds write access to $CORPUS and swaps a validated ancestor for an external symlink between the
+#   check and the mkdir/mv. The gateway lock does NOT bind such a process. This is unsolvable at the
+#   application layer (a process with arbitrary write to $CORPUS can `rm -rf $CORPUS` outright); the
+#   real control is OS-level filesystem ownership/permissions — the corpus dir must be owned/writable
+#   ONLY by the gateway's uid. Deploy the corpus that way; these checks are the in-process complement,
+#   not a substitute for the filesystem-ownership invariant.
 _assert_within_corpus() {  # _assert_within_corpus <dest-path> — 0 iff the resolved parent dir stays inside canonical CORPUS
   local dir real                                   # (CORPUS is already `pwd -P` canonical; realpath here defeats a symlinked type-dir escape)
   dir="$(dirname "$1")"
--- a/bin/memory-gateway
+++ b/bin/memory-gateway
@@ -64,9 +64,21 @@ set -uo pipefail
 # ---------------------------------------------------------------------------
 # setup / tool probes
 # ---------------------------------------------------------------------------
-JQ="$(command -v jq || true)"
+# Binary resolution: PATH first (command -v), then known absolute candidates. launchd/cron strip PATH to a
+# minimal set, and this gateway is designed for UNATTENDED/hook invocation, so a bare `command -v` would make
+# it spuriously fail there. Candidates cover macOS Homebrew (arm+intel), system, and common Linux locations.
+_resolve_bin() {  # _resolve_bin <name> <abs-candidate>...  -> prints an absolute path, or nothing (rc 1)
+  local p; p="$(command -v "$1" 2>/dev/null || true)"; [ -n "$p" ] && { printf '%s' "$p"; return 0; }
+  shift; for c in "$@"; do [ -x "$c" ] && { printf '%s' "$c"; return 0; }; done
+  return 1
+}
+# version-manager shims (mise/asdf) are version-independent absolute paths that work under an empty PATH, so
+# they cover managed installs the fixed system paths miss; ${HOME:-} guards an unset HOME under set -u.
+_MISE="${HOME:-}/.local/share/mise/shims"; _ASDF="${HOME:-}/.asdf/shims"
+JQ="$(_resolve_bin jq "$_MISE/jq" "$_ASDF/jq" /opt/homebrew/bin/jq /usr/local/bin/jq /usr/bin/jq /bin/jq || true)"
 [ -n "$JQ" ] || { echo "memory-gateway: jq is required (install jq)" >&2; exit 2; }
-PYTHON3="$(command -v python3 || true)"
+PYTHON3="$(_resolve_bin python3 "$_MISE/python3" "$_ASDF/python3" /opt/homebrew/bin/python3 /usr/local/bin/python3 /usr/bin/python3 /bin/python3 || true)"
+PS_BIN="$(_resolve_bin ps /bin/ps /usr/bin/ps || true)"   # for _lock_is_stale liveness disambiguation (EPERM vs ESRCH); ps is a system binary, no shim
 
 VERB=""
 CORPUS=""
@@ -114,6 +126,12 @@ _err_setup() {  # _err_setup <message>   (setup/environment failure)
   _emit error "${VERB:-}" "" null "$("$JQ" -nc --arg c SETUP --arg m "$1" '{code:$c, message:$m}')"
   exit 2
 }
+_err_io() {  # _err_io <message>  — mid-op I/O failure (disk full / perms): the environment cannot persist.
+             # exit 2 (environment-can't-serve, [C06]) — NEVER a false `ok`. In a WAL transaction the intent
+             # stays open, so the next invocation's recovery completes the mutation idempotently.
+  _emit error "${VERB:-}" "" null "$("$JQ" -nc --arg c IO_ERROR --arg m "$1" '{code:$c, message:$m}')"
+  exit 2
+}
 
 # ---------------------------------------------------------------------------
 # small helpers
@@ -146,6 +164,14 @@ _valid_slug() {  # safe type/name slug; rejects traversal / absolute / illegal
   [ -n "$type" ] && [ -n "$name" ] || return 1
   _valid_type "$type" || return 1
   return 0
+}
+
+_NL="$(printf '\nx')"; _NL="${_NL%x}"   # one literal LF (the `printf x`+strip guard survives command-subst)
+_CR="$(printf '\rx')"; _CR="${_CR%x}"   # one literal CR
+_no_ctrl_lines() {  # _no_ctrl_lines <string> — 0 iff no LF/CR. Frontmatter fields (name/description) are
+                    # single-line by construction; an embedded newline would inject/terminate the `---` block
+                    # (identity spoof / structural corruption). Reject at input; multi-line detail belongs in the body.
+  case "$1" in *"$_NL"*|*"$_CR"*) return 1 ;; *) return 0 ;; esac
 }
 
 _fm_field() {  # _fm_field <file> <field>  -> value of a frontmatter field
@@ -195,18 +221,26 @@ _assert_within_corpus() {  # _assert_within_corpus <dest-path> — 0 iff the res
   case "$real" in "$CORPUS"|"$CORPUS"/*) return 0 ;; *) return 1 ;; esac
 }
 
+_assert_read_safe() {  # _assert_read_safe <path> — read-side symmetry of the write BOUNDARY guard: the resolved
+                       # parent stays inside CORPUS AND the target itself is not a symlink escaping it. read/search/
+                       # neighborhood trusted _valid_slug (string-safe) but not this realpath check -> a symlinked
+                       # type-dir (or symlinked topic file) could exfiltrate outside the corpus. defense-in-depth.
+  _assert_within_corpus "$1" && ! [ -L "$1" ]
+}
+
 _atomic_write() {  # _atomic_write <dest> <content>   (content is an ARG, NOT stdin: this MUST run in the
                    # caller's shell so a boundary _refuse can abort the whole process — a pipe-subshell exit
                    # cannot, which silently produced a double-envelope + exit 0 on a symlink escape; temp->rename+fsync)
   local dest="$1" content="$2" dir tmp
   dir="$(dirname "$dest")"
-  mkdir -p "$dir"
+  mkdir -p "$dir" || return 1
   _assert_within_corpus "$dest" || _refuse "${VERB:-}" "" BOUNDARY "resolved write path escapes the corpus root (symlinked dir?): $dest"
   tmp="$(mktemp "$dir/.mgw.XXXXXX")" || return 1
-  printf '%s' "$content" > "$tmp"
+  if ! printf '%s' "$content" > "$tmp"; then rm -f "$tmp" 2>/dev/null || true; return 1; fi  # e.g. ENOSPC/quota -> NEVER false-ok
   _fsync "$tmp"
-  mv "$tmp" "$dest"
+  mv "$tmp" "$dest" || { rm -f "$tmp" 2>/dev/null || true; return 1; }
   _fsync "$dir"
+  return 0
 }
 
 # ---------------------------------------------------------------------------
@@ -215,7 +249,16 @@ _atomic_write() {  # _atomic_write <dest> <content>   (content is an ARG, NOT st
 _lock_is_stale() {  # _lock_is_stale <lockdir> — stale if the holder PID is provably dead OR ts older than LOCK_STALE_SECS
   local ts now age pid
   pid="$(cat "$1/pid" 2>/dev/null || true)"
-  case "$pid" in ''|*[!0-9]*) ;; *) kill -0 "$pid" 2>/dev/null || return 0 ;; esac  # holder dead -> reclaim NOW (real-crash recovery, not 900s wait)
+  # `kill -0` returns non-zero for BOTH ESRCH (dead -> reclaim) AND EPERM (holder ALIVE but owned by another
+  # UID -> must NOT reclaim; blind reclaim would evict a live cross-UID writer -> concurrent writers). Disambiguate
+  # with `ps -p` (UID- AND locale-independent, unlike stderr text): only a pid ps ALSO cannot see is truly dead ->
+  # fast real-crash reclaim; a live-but-other-UID holder (or ps unavailable) falls through to the ts backstop
+  # below (fail-safe: never evict a holder we cannot PROVE is dead).
+  case "$pid" in ''|*[!0-9]*) ;; *)
+    if ! kill -0 "$pid" 2>/dev/null; then
+      if [ -n "$PS_BIN" ]; then "$PS_BIN" -p "$pid" >/dev/null 2>&1 || return 0; fi
+    fi ;;
+  esac
   ts="$(cat "$1/ts" 2>/dev/null || true)"
   case "$ts" in ''|*[!0-9]*) return 1 ;; esac  # no valid ts yet (just-acquired window) -> NOT stale (fail-safe: never reclaim a lock you can't prove is old)
   now="$(date +%s)"; age=$(( now - ts ))
@@ -226,7 +269,13 @@ _lock_acquire() {  # refuses CONCURRENT_WRITER on timeout (exit 1)
   LOCK="$CORPUS/.gateway/lock.d"
   local tries=0
   while ! mkdir "$LOCK" 2>/dev/null; do
-    if _lock_is_stale "$LOCK"; then rm -rf "$LOCK" 2>/dev/null || true; continue; fi
+    if _lock_is_stale "$LOCK"; then
+      # single-winner reclaim: atomically RENAME the stale dir out of the way; only ONE racer wins moving the
+      # ORIGINAL (a concurrent reclaimer's mv finds the source already gone -> fails harmlessly). blind `rm -rf`
+      # + retry had a TOCTOU: two racers could both delete-then-mkdir -> two live writers. rename is atomic.
+      if mv "$LOCK" "$LOCK.stale.$$" 2>/dev/null; then rm -rf "$LOCK.stale.$$" 2>/dev/null || true; fi
+      continue
+    fi
     tries=$(( tries + 1 ))
     [ "$tries" -ge "$LOCK_MAX_TRIES" ] && _refuse "$VERB" "" CONCURRENT_WRITER \
       "another writer holds the corpus lock ($LOCK) — refusing to clobber; retry shortly"
@@ -262,18 +311,22 @@ _crash_if() {  # _crash_if <stage>
 # external + env-overridable (matches artifact-registry), so tests isolate it.
 # ---------------------------------------------------------------------------
 _audit() {  # _audit <verb> <slug>
-  local dir led llock i=0
+  local dir led llock i=0 acquired=0
   dir="${MEMORY_GATEWAY_DIR:-$HOME/.claude/audit}"
   mkdir -p "$dir"
   led="$dir/memory-gateway.jsonl"
   llock="$dir/.lock-memory-gateway"
-  while ! mkdir "$llock" 2>/dev/null; do
-    i=$(( i + 1 )); [ "$i" -ge 50 ] && { rm -rf "$llock" 2>/dev/null || true; break; }
+  while true; do
+    if mkdir "$llock" 2>/dev/null; then acquired=1; break; fi
+    i=$(( i + 1 ))
+    [ "$i" -ge 50 ] && break   # give up the lock: do NOT force-remove a LIVE holder's dir (that clobber could
+                               # interleave two appenders). the append below is ONE sub-PIPE_BUF line -> atomic
+                               # under POSIX even unlocked, so proceeding lock-free is safe (liveness > exclusivity here).
     sleep 0.02
   done
   "$JQ" -nc --arg ts "$(_ts)" --arg verb "$1" --arg slug "$2" --arg corpus "$CORPUS" \
     '{ts:$ts, verb:$verb, slug:(if $slug=="" then null else $slug end), corpus:$corpus}' >> "$led"
-  rm -rf "$llock" 2>/dev/null || true
+  [ "$acquired" = 1 ] && { rm -rf "$llock" 2>/dev/null || true; }   # release ONLY the lock WE created
 }
 
 # ---------------------------------------------------------------------------
@@ -295,7 +348,7 @@ _rebuild_index() {  # rebuilds MEMORY.md; echoes the live-entry count
   done
   local _idx
   _idx="$( { printf '# Memory Index\n\n> Regenerated by memory-gateway (ADR-013). One line per live topic.\n\n'; printf '%s' "$body"; }; printf x )"; _idx="${_idx%x}"
-  _atomic_write "$CORPUS/MEMORY.md" "$_idx"
+  _atomic_write "$CORPUS/MEMORY.md" "$_idx" || _err_io "failed to persist the index (MEMORY.md) — I/O error"
   REBUILD_COUNT="$lines"   # global out-param: lets callers read the count WITHOUT $() — a $()
                            # capture would re-nest _atomic_write in a command-substitution
                            # subshell, where its BOUNDARY _refuse could only kill the subshell
@@ -315,7 +368,7 @@ _archive_topic() {  # _archive_topic <slug>  -> moves live to archive, echoes se
   seq=$(( cnt + 1 ))
   padded="$(printf '%04d' "$seq")"
   dst="$adir/$padded.md"
-  mv "$live" "$dst"
+  mv "$live" "$dst" || _err_io "failed to move '$slug' into the archive tier — I/O error (refusing to proceed; the superseding write must never overwrite an un-archived fact)"
   _fsync "$(dirname "$live")"; _fsync "$adir"
   printf '%s' "$seq"
 }
@@ -370,7 +423,7 @@ _replay_supersede() {  # _replay_supersede <intent-json>  (idempotent). returns 
     fi
   fi
   [ -n "$superseded" ] && [ "$superseded" != "null" ] && _archive_if_stale "$superseded" "$payload" >/dev/null
-  _atomic_write "$CORPUS/$new_slug.md" "$payload"
+  _atomic_write "$CORPUS/$new_slug.md" "$payload" || _err_io "recovery: failed to write '$new_slug' — I/O error (intent left OPEN for the next recovery; no commit marker written)"
   _rebuild_index >/dev/null
 }
 
@@ -418,7 +471,7 @@ _wal_supersede() {  # _wal_supersede <new_slug> <superseded_slug> <content>
   archseq="$(_archive_if_stale "$superseded" "$content")"
   _crash_if archive
   # STAGE 3 — write the new live topic (atomic temp->rename, fsync)
-  _atomic_write "$CORPUS/$new_slug.md" "$content"
+  _atomic_write "$CORPUS/$new_slug.md" "$content" || _err_io "supersede: failed to write new topic '$new_slug' — I/O error (WAL intent stays OPEN; recovery completes it idempotently)"
   _crash_if newtopic
   # STAGE 4 — rebuild the index
   _rebuild_index >/dev/null
@@ -472,6 +525,8 @@ do_create() {
   [ -n "$TYPE" ] || _refuse create "" USAGE "create requires --type {user|feedback|project|reference}"
   _valid_type "$TYPE" || _refuse create "" USAGE "invalid --type '$TYPE'"
   [ -n "$NAME" ] || _refuse create "" USAGE "create requires a topic name (positional)"
+  _no_ctrl_lines "$NAME" || _refuse create "" USAGE "name must be a single line (no newline/CR — would inject into the frontmatter)"
+  _no_ctrl_lines "$DESC" || _refuse create "" USAGE "description must be a single line (no newline/CR); put multi-line detail in the body"
   norm="$(_normalize "$NAME")"
   [ -n "$norm" ] || _refuse create "" USAGE "name '$NAME' normalizes to empty"
   slug="$TYPE/$norm"
@@ -483,7 +538,7 @@ do_create() {
   fi
   body="$STDIN_BODY"   # pre-read off-lock in dispatch (#1); byte-identical to the old inline $(cat)
   content="$(_topic_content "$NAME" "$TYPE" "$DESC" "$body")"
-  _atomic_write "$path" "$content"
+  _atomic_write "$path" "$content" || _err_io "create: failed to persist '$slug' — I/O error (no false success reported)"
   _rebuild_index >/dev/null
   _audit create "$slug"
   _emit ok create "$slug" \
@@ -508,6 +563,7 @@ do_read() {
     if [ -d "$adir" ]; then
       for f in "$adir"/*.md; do
         [ -f "$f" ] || continue
+        _assert_read_safe "$f" || continue   # skip a symlink-escaping archived file (defense-in-depth)
         seq="$(basename "$f" .md)"; rel="${f#$CORPUS/}"; c="$(cat "$f")"
         archived_json="$("$JQ" -nc --argjson a "$archived_json" --arg seq "$seq" --arg path "$rel" --arg content "$c" \
           '$a + [{seq:$seq, path:$path, content:$content}]')"
@@ -519,6 +575,7 @@ do_read() {
   else
     path="$CORPUS/$SLUG.md"
     [ -f "$path" ] || _refuse read "$SLUG" NOT_FOUND "no live topic at slug '$SLUG' (try --archived)"
+    _assert_read_safe "$path" || _refuse read "$SLUG" BOUNDARY "resolved read path escapes the corpus (symlinked dir/file?)"
     content="$(cat "$path")"
     _emit ok read "$SLUG" "$("$JQ" -nc --arg s "$SLUG" --arg c "$content" '{slug:$s, content:$c}')" null
   fi
@@ -545,7 +602,7 @@ _update_append() {
   delta="$STDIN_BODY"   # pre-read off-lock in dispatch (#1); byte-identical to the old inline $(cat)
   local _merged
   _merged="$( { cat "$path"; printf '\n%s\n' "$delta"; }; printf x )"; _merged="${_merged%x}"
-  _atomic_write "$path" "$_merged"
+  _atomic_write "$path" "$_merged" || _err_io "update: failed to persist append to '$SLUG' — I/O error (no false success reported)"
   _audit update "$SLUG"
   _emit ok update "$SLUG" "$("$JQ" -nc --arg s "$SLUG" '{slug:$s, appended:true}')" null
 }
@@ -570,6 +627,8 @@ _update_supersede() {
   [ -n "$TYPE" ] || _refuse update "" USAGE "supersede requires --type"
   _valid_type "$TYPE" || _refuse update "" USAGE "invalid --type '$TYPE'"
   [ -n "$NAME" ] || _refuse update "" USAGE "supersede requires --name"
+  _no_ctrl_lines "$NAME" || _refuse update "" USAGE "name must be a single line (no newline/CR — would inject into the frontmatter)"
+  _no_ctrl_lines "$DESC" || _refuse update "" USAGE "description must be a single line (no newline/CR); put multi-line detail in the body"
   norm="$(_normalize "$NAME")"
   [ -n "$norm" ] || _refuse update "" USAGE "name '$NAME' normalizes to empty"
   new_slug="$TYPE/$norm"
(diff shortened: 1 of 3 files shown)


ekson73:
@qodo-code-review

Thanks — worked all 8 findings against the real code (verified each, since bots false-positive too). **6 fixed with regression pins, 2 deferred by-design** (none dropped). Fixes in `850bbbc`; suite is now 57 assertions, 20/20 deterministic.

| # | Finding | Disposition | Fix + pin |
|---|---|---|---|
| bin:234 | Reclaim lock could delete the owner (TOCTOU) | ✅ Fixed | `_lock_acquire` steals via **atomic `mv "$LOCK" "$LOCK.stale.$$"`** — renames the dir it observed, never `rm`s one a peer may have reacquired. Pin: test 37. |
| bin:523 | Read ignores the boundary symlink | ✅ Fixed | `_assert_read_safe` (realpath-within-root **and** not-a-symlink) on `read`/`search`/`neighborhood`. Pin: test 39. |
| bin:210 | I/O failures masked as `ok` | ✅ Fixed | `_err_io` → **exit-2 `IO_ERROR`** (never a false `ok`); every mutation `\|\| return 1` / `\|\| _err_io`. The open WAL intent is completed idempotently by the next invocation's sha256-guarded recovery. Pin: test 39. |
| bin:167 | Frontmatter accepts newline injection | ✅ Fixed | `_no_ctrl_lines` rejects `\n`/`\r` in `--name`/`--description` (`USAGE`). Pin: test 40. |
| tests:233 | E6 test misses `neighborhood` | ✅ Fixed (code + test) | Real gap too: `do_neighborhood` gained an E6 liveness gate `[ -f "$CORPUS/$l.md" ] \|\| continue` so a superseded slug is never a neighbor; new test 38 asserts it. |
| bin:69 | jq path lacks fallbacks | ✅ Fixed | `_resolve_bin` (`command -v` → absolute candidates incl. mise/asdf shims). Pin: test 42 (resolves + runs under empty PATH). |
| bin:28 | `purge` verb not implemented | ⏭️ Deferred (by-design) | ADR-013 non-goal: forgetting = archive-tier supersession (E6), retrievable via `.archive/`. A compliant hard-erasure/purge is a distinct future ADR item. |
| bin:167 | `_topic_content` lacks bi-temporal fields | ⏭️ Deferred (by-design) | ADR-013 chose archive-tier + WAL supersession over `valid_until`/`superseded_by` frontmatter. E6-forgetting **is** enforced (archive move + liveness gates), just not via metadata. Alternative, not a defect. |

Note: the SkillSpector MP2 "context-window stuffing" sub-warning on SKILL.md is a false-positive — the edit *removed* an over-broad exception (net +2 lines).

Could you please re-review?

ekson73:
@coderabbitai

Thanks — 2 fixed with pins, 1 deferred with rationale (verified each against the code). Fixes in `850bbbc`; suite is now 57 assertions, 20/20 deterministic.

| # | Finding | Disposition | Fix + pin |
|---|---|---|---|
| bin:218 | EPERM ≠ ESRCH in `_lock_is_stale` | ✅ Fixed | `kill -0` is non-zero for a dead pid **and** a live pid on another UID → a `ps -p` probe now disambiguates: only a pid `ps` **also** can't see is truly dead (fast reclaim); a live-other-UID holder (or `ps` unavailable) falls through to the ts-age backstop — **fail-toward-safe** (never evict an unproven-dead holder → no double-writer on a canonical mutator). Pin: test 41. |
| SKILL.md:103 | Unrestricted direct edits | ✅ Fixed | Advisory Posture tightened: identity-change/supersede are **gateway-only**; a direct edit is tolerated only as a last-resort **quiesced** migration + `index` reindex + re-read verification. |
| bin:357 | WAL never compacted | ⏭️ Deferred | Agreed it's a nitpick ("Correctness is unaffected"). Compaction must coordinate with the corpus lock to avoid racing a concurrent appender — mis-done it risks the durability the WAL provides. Steady-state-only cost on a long-lived corpus → deferred with a tracked note (premature optimization). |

Could you please re-review?

ekson73:
@coderabbitai

Both findings on `850bbbc` verified against the real code and **fixed** in `4af469a`; the WAL nitpick re-affirmed as a tracked defer.

| Finding | Verdict | Fix |
|---|---|---|
| **[763] Major** — neighborhood surfaces a symlink-escaping neighbor slug (`[ -f ]` follows the link; a leaf neighbor is never promoted to `$node` so the line-763 node-guard can't catch it) | ✅ genuine | Applied the same already-verified `_assert_read_safe` at neighbor-discovery (`bin/memory-gateway:769`) — `read`/`search`/`neighborhood` now share ONE realpath boundary over every surfaced item, and the escaping slug is also kept out of the frontier. |
| **[606/626] Minor · SC2015** — `A && B \|\| C` in tests 40/41 lets `no` double-fire | ✅ genuine | Converted both to explicit `if/then/else`. |
| **[349] WAL never compacted** (🔵 Trivial, "Correctness unaffected") | ⏸️ defer | Compaction must coordinate with the corpus lock to avoid racing a concurrent appender; steady-state-only cost → premature optimization (measure-first). Tracked in the PR body's Deferred section. |

**Regression pin — test 43** (proven non-tautological): seeds a node linking to a symlink-escaping slug with a **real** outside target (so the E6 `[ -f ]` liveness gate passes and the new guard is actually exercised) + a legit in-corpus positive control. It **FAILS without the guard** (escaped slug leaks into `nodes`) and **PASSES with it** — verified via overlay-and-restore against the real path (SHA-checked restore).

Tests: **58/58 passing** ✅

Could you please re-review?

ekson73:
@qodo-code-review — thank you, you were right: finding **[189] (I/O failures masked)** was **not** fully closed. Fixed in `72477ad`.

**The hole you found (verified genuine in the real code):** in `_wal_supersede` STAGE 2, `archseq="$(_archive_if_stale …)"` ran the archive inside a command-substitution subshell, so a failed archive-`mv` → `_err_io`'s `exit 2` killed only the subshell; the parent continued, wrote the successor **over the un-archived original**, stamped the WAL commit marker, and emitted `ok` — breaking both E6 atomicity and the intent-stays-open recovery guarantee. On the pre-fix code the injected failure produced `r=0 status=ok` with the original clobbered.

**Root fix (your suggested approach — capture without command substitution):** `_archive_topic` / `_archive_if_stale` now set a global `$ARCHSEQ` instead of `printf`-ing the seq for `$()` to read. They run in the **caller's** process, so `_err_io`'s `exit 2` propagates to the top-level error envelope. One edit fixes **all three** affected sites — not just the one you flagged:

| Site | Was | Now |
|---|---|---|
| `bin/memory-gateway:471` `_wal_supersede` STAGE 2 (flagged) | `archseq="$(_archive_if_stale …)"` | `_archive_if_stale …; archseq="$ARCHSEQ"` |
| `bin/memory-gateway:693` `do_archive` verb (**same latent bug**, uncaught by the line-471 view) | `seq="$(_archive_topic …)"` | `_archive_topic …; seq="$ARCHSEQ"` |
| `bin/memory-gateway:425` `_replay_supersede` recovery | `_archive_if_stale … >/dev/null` (also **ate the `[C06]` error envelope**) | `_archive_if_stale …` (envelope now reaches real stdout, intent left OPEN) |

**Regression coverage for the archive-MOVE path** (the gap you named — test 39 only exercised a topic-WRITE failure): new **test 44** injects an archive-move `EACCES` during a supersession (target `.archive` dir `chmod 555`) → asserts exit-2 `IO_ERROR` + the original still live (successor never overwrote it). I verified it's a **non-tautological** pin: it **fails on the faithful pre-fix code** (`git HEAD:bin/memory-gateway`) with **only test 44 red** (`r44=0 s44=ok orig_kept=0`) and passes on the fix — zero collateral.

**Bookkeeping note — corrected:** you were right that the pass-4 reply table cited stale pin numbers. Current numbering: symlink-escape write-boundary = **31** · newline injection = **37** · stale-lock reclaim = **40/41** · read-side symlink-neighbor exclusion = **43** · archive-move I/O failure = **44**.

Tests: **59/59** passing. gitleaks clean. Could you please re-review `72477ad`?

ekson73:
@coderabbitai

Addressed the outside-diff **Major** finding on `_archive_topic` (`bin/memory-gateway:363-375`) in `61439a4`.

| Finding | Verification (real code, Mente Tomé) | Fix | Commit |
|---|---|---|---|
| A symlinked `.archive` can send `mv` outside `CORPUS` | **Confirmed** — `_atomic_write` guards the topic-WRITE path with `_assert_within_corpus` before its `mv` (test 31), but `_archive_topic`'s `mv "$live" "$dst"` did not; a pre-planted `.archive` symlink escapes the corpus root (same vector tests 31/43 defend). | `_assert_within_corpus "$dst" \|\| _refuse … BOUNDARY` before the `mv` (mirrors `_atomic_write` exactly). | `61439a4` |
| Bare `mkdir -p` can exit without an `_err_io` envelope on a filesystem failure | **Confirmed** — `_atomic_write` fails-fast (`\|\| return 1`); `_archive_topic` did not. | `\|\| _err_io` on the archive `mkdir -p`. | `61439a4` |

**Regression pin — test 45** (`CodeRabbit#PDCA-7`): symlinks `.archive` → outside, runs `archive`, asserts refuse `BOUNDARY` + live topic un-moved + **no `*.md` content written outside** (mkdir may leave an empty out-of-corpus dir exactly as `_atomic_write` does — the guarantee is that no content leaves, parallel to test 31's `! -e $OUTSIDE/pwned.md`). **Non-tautology proven**: against the faithful pre-fix code (HEAD `72477ad`) **only test 45** goes red (`r45=0 c45=null kept45=0 nooutside45=0 live=[]`); the fix makes it green with zero collateral.

Tests: **60 assertions passing** ✅ · gitleaks clean on both changed files ✅

This applies the already-red-teamed boundary primitive (tests 31/43) to one more call site — it hardens an existing failure path, no new attack surface.

Could you please re-review `61439a4`?

ekson73:
## PDCA round 8 — CodeRabbit actionable fixed (`e9beb98`) + qodo whole-file findings triaged

### @coderabbitai — PDCA-7 re-review actionable addressed

Your re-review of `61439a4` asked to validate the archive path **before any filesystem change** so no artifact is created outside `CORPUS` on a symlink escape, and to strengthen the regression test. Applied in `e9beb98`:

| Your ask | Fix | Where |
|---|---|---|
| Refuse symlinked `.archive` **before** `mkdir -p` (no artifact created) | New helper `_assert_new_path_within_corpus` resolves the **deepest existing ancestor** of the not-yet-created `$adir` and asserts it's inside `CORPUS`; called **before** `mkdir` in `_archive_topic` (replaces the post-mkdir `_assert_within_corpus "$dst"`). A `mkdir -p` path can only escape through a symlinked *existing* ancestor (the created tail is literal/symlink-free), so the existing-prefix check is necessary & sufficient — zero fs artifact. | `bin/memory-gateway` `_assert_new_path_within_corpus` + `_archive_topic` |
| Run the regression test **as root too** | Removed the `if [ "$(id -u)" != 0 ]` skip — the symlink boundary is UID-independent (`cd`+`pwd -P`), unlike test 44 which relies on a `chmod 555` perm-bypass that root defeats. | `tests/test-memory-gateway.sh` test 45 |
| Assert the outside dir has **no descendants**, not only no `*.md` | `find "$ASOUT" -mindepth 1` must be empty — **ZERO** artifacts, which now holds precisely because the refusal precedes `mkdir`. | test 45 |

**Non-tautology proof:** overlaying the pre-refinement bin (`61439a4`, post-mkdir check) against the *strengthened* test 45 turns **exactly one** assertion red (`noart45=0` — the post-mkdir version leaves an empty escaped dir), with `r45=1 / c45=BOUNDARY / kept45=1` still holding and 59 others green. The before-mkdir change is genuinely necessary, not decorative.

**Deferred follow-up (transparency):** `_archive_topic` is now *stricter* than `_atomic_write` (which still checks post-mkdir). Hardening `_atomic_write` to the same before-mkdir invariant is a sound consistency follow-up, but I've **kept it out of this PR** to avoid touching the already-red-teamed write path (tests 31/44 pin it); it's noted as a deferred item, not silently dropped.

60/60 assertions pass; gitleaks clean on both files. Could you please re-review `e9beb98`?

---

### @qodo-code-review — 8 findings triaged (verified against the current code)

I verified every finding against the real code + test suite (Mente Tomé — not just titles). Each is either out-of-scope **by ADR-013 design** or already addressed with a named pinning test. Details:

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | `purge` verb not implemented | **Out-of-scope by design** | ADR-013 line 37: *"Purge / true-erasure is deliberately OUT of scope (tiering-never-deletion)."* MVP has no production PII (`vek-mvp-env-scope`); `archive` is the only removal verb (content never destroyed). Operator-ratified. |
| 2 | `_topic_content` missing bi-temporal fields (`valid_until`/`superseded_by`) | **Different ratified design** | ADR-013 §E6 specifies **archive-based** forgetting: a superseded fact **leaves** the active corpus (tiered into `.archive/`) rather than co-residing live with a `superseded_by` pointer. The frontmatter (`name`/`type`/`description`) is correct-by-spec; bi-temporal metadata is an alternative model, not an ADR-013 requirement. |
| 3 | E6 test misses neighborhood | **Already tested** | `test 38` ("superseded neighbor ABSENT from get_neighborhood — present before, gone after supersession"). |
| 4 | `jq` path lacks fallbacks | **Already implemented** | `_resolve_bin` (PATH → mise/asdf shims → homebrew arm+intel → system paths); `test 42` is literally named `qodo#4` and pins jq resolution + run under an **empty PATH**. |
| 5 | Reclaim lock pode apagar o dono (**Bug**) | **False positive** | `_lock_acquire` reclaims via **atomic `mv "$LOCK" "$LOCK.stale.$$"`** rename-steal (only one racer wins moving the original; the loser's `mv` finds the source gone and retries `mkdir`). The subsequent `rm -rf` targets **only** the PID-quarantined `$LOCK.stale.$$` name — never the live lock. `_lock_is_stale` disambiguates ESRCH (dead→reclaim) vs EPERM (live cross-UID→never reclaim) via `ps`. Pinned by `tests 40/41`. |
| 6 | Leitura ignora boundary symlink (**Bug/Security**) | **Already enforced** | `_assert_read_safe` (`_assert_within_corpus && ! [ -L ]`) gates **every** read path: `do_read` (refuse `BOUNDARY` at line 594; archived-skip at 582) and `do_neighborhood` (skip at 779/785). Pinned by `test 43` ("read/search/neighborhood share one boundary"). The flagged line 596 *is* the refusal line. |
| 7 | Falhas de I/O mascaradas (**Bug**) | **Already handled** | Every mutating op checks `\|\| _err_io` (index 363, archive-dir 379, archive-move 384, recovery 441, supersede-write 490). The `$ARCHSEQ` **global out-var** (not `$()`) pattern exists specifically so `_err_io`'s `exit 2` propagates instead of being swallowed by a command-substitution subshell. Pinned by `test 39` (write I/O → exit 2 IO_ERROR) + `test 44` (archive-move I/O → exit 2). |
| 8 | Frontmatter aceita injeção newline (**Bug**) | **Already refused** | `_no_ctrl_lines` rejects newline/CR in `name`/`description` with `USAGE` (create: lines 544/545; supersede: 646) — no frontmatter injection, nothing written. Pinned by `test 37`. |

Net: findings #1–#2 are ADR-013-ratified scope decisions; #3–#8 were addressed in prior PDCA rounds and are pinned by named tests (37/38/39/40/41/42/43/44). No code change required for these — happy to discuss any you'd like to re-examine.

<sub>60/60 tests pass · gitleaks clean · HEAD `e9beb98` · ADR-013 · EKO-116</sub>

ekson73:
@coderabbitai — thanks for confirming the before-mkdir archive fix. On your remaining TOCTOU caveat:

You're right that `_assert_new_path_within_corpus` (like any in-process realpath check) is a check-then-use, so a **non-cooperating** process holding write access to `$CORPUS` could swap a validated ancestor between the check and the `mkdir`/`mv`. That race is **out of scope by design and unclosable at this layer** — a process with arbitrary write to the corpus can `rm -rf $CORPUS` outright, so the only real control is OS-level filesystem ownership. Per your suggestion, I made the invariant **explicit** rather than implicit (`474bd19`):

| Where | Note |
|---|---|
| `bin/memory-gateway` | New **`BOUNDARY THREAT MODEL`** comment block above the three boundary helpers: in-scope = a **pre-planted** symlink under the gateway's ADR-013 single-writer contract (the check is exact there); out-of-scope = a **concurrent non-cooperating** FS mutator (the corpus lock does not bind it) — closed only by the deploy invariant that `$CORPUS` is owned/writable **only** by the gateway's uid. |
| `skills/memory-gateway/SKILL.md` | The **Boundary-safe** guarantee now names the same single-writer / FS-ownership invariant. |

Documentation only — no logic change; 60/60 assertions pass, gitleaks clean. The filesystem-ownership invariant is now the explicit contract you asked for. Could you please re-review `474bd19`?

ekson73:
## ✅ Merging under standing auto-merge authorization — convergence verified

Per `auto-merge-standing-authorization.md` (standing HITL grant 2026-05-25) + `pr-review-protocol.md` §2.6.1 (G1-G8 / R1-R6). The 3 gates hold on HEAD `474bd19`:

| Gate | Status | Evidence |
|---|---|---|
| **C1 — high-score-confident** | ✅ ≥0.90 | Every line authored + verified; 60/60 assertions pass; each fix carries a **non-tautology proof** (pre-fix bin overlaid against the strengthened test turns exactly one pin red, zero collateral) |
| **C2 — green PR** | ✅ | `mergeable=MERGEABLE`; 8 check-runs SUCCESS/neutral (gitleaks · governance-validation · Trivy ×2 · Scorecard ×2 · pip-audit · Lockfile); `CodeRabbit`=success · `security/snyk`=success. Only `code/snyk`=error = **known Snyk quota non-blocker** ("Code test limit reached"), and the main ruleset has **no `required_status_checks` rule** → advisory only (state=UNSTABLE, not BLOCKED) |
| **C3 — agentic convergence** | ✅ | **CodeRabbit cleared**: *"This resolves the remaining TOCTOU documentation caveat; no further concern. 🐇"* (comment 5064407715, on `474bd19`). **qodo**: 8 findings all COMMENTED/advisory — each rebutted with evidence (out-of-scope-by-ADR-013 `purge`/bi-temporal, or already-pinned by named tests 37-44). 3 stale CodeRabbit CHANGES_REQUESTED reviews (on superseded commits 4d64db6/850bbbc/61439a4) dismissed per §8 |

**Deny-set clear** (no HUMAN_DOMAIN): synthetic `bin/` + `tests/` fixtures only — no secrets (gitleaks clean), no PII, no production data. Merge to `main` on a plugin repo (GitHub Flow, no deploy-on-merge) is git-revertible — not irreversible. Anti-theater 8/8 (real fix, verified, non-tautology-proven).

**PDCA journey**: 8 rounds. CodeRabbit's final genuine actionable (validate the archive path **before** `mkdir`, so a symlinked `.archive` escape creates zero fs artifact) → fixed via `_assert_new_path_within_corpus` (`e9beb98`) → CodeRabbit confirmed correct + raised a TOCTOU documentation caveat → threat-model doc added (`474bd19`, doc-only) distinguishing in-scope pre-planted symlinks from the out-of-scope non-cooperating-FS-mutator (an OS-level filesystem-ownership concern: `$CORPUS` owned/writable only by the gateway uid) → CodeRabbit cleared.

Squash-merging with `--delete-branch`. Follow-up (deferred, tracked in PR body): harden `_atomic_write` to the same before-mkdir invariant for consistency (kept out of this PR to avoid touching the already-red-teamed write path pinned by tests 31/44).

<sub>Commits unsigned per `multi-identity-git-ssh-governance` §6.1 (1P SSH-signing agent down; signature-only pre-authorization — hooks ran, gitleaks clean). ADR-013 · EKO-116.</sub>

PR #288, opened 2026-07-29
feat(skills): research-dossier — provenance-gated research→dossier IR (#287)

### LINKED ISSUE #287: feat(skills): research-dossier — research corpus to decision-ready visual dossier
## Context

Research output and dashboard renderers are both well covered in this ecosystem, but they do not connect. Researchers emit prose markdown; every good renderer wants structured input (`data-report` ← csv/xlsx/json, `d3-visualization` ← arrays, `session-report` ← analyzer JSON). Nothing converts one into the other, so every dashboard is hand-rolled per session, with no provenance and no reuse of the ~110 `design-templates/` and ~150 `design-systems/` already installed via the `open-design` marketplace.

Three concrete gaps:

1. **No question → data adapter.** The gap that sustains the other two.
2. **No comparison primitive.** "Compare x, y, z" needs a normalized N-options × M-criteria matrix with evidence, confidence, and source *per cell*.
3. **No 1-source → N-format fan-out.** `content-recast` picks one `--format` per invocation; `graphify` emits three but is hard-wired to graphs. Nothing emits html+pdf+pptx+json from one intermediate representation with a consistent theme.

Plus two inherited defects this closes at the root:

4. **The faithfulness gate does not survive into visuals.** `content-recast`'s gate operates on *text*. Nothing verifies that a truncated axis distorts a magnitude — even though "no magnitude distortion" is rule 3 of that same gate.
5. **`--audience` never reaches the design system.** Audience only selects a lens; there is no `ceo → publication`, `engineer → mission-control` map.

## Objective

A thin orchestrator plus three internal primitives that route `corpus → IR → (scorecard) → render`. The intermediate representation is the deliverable, not the HTML: once research becomes JSON with per-claim provenance, all existing templates and renderers become reachable for free.

## Approach

**Compose, never duplicate.** Zero new design systems (~150 exist). Zero new per-format renderers (6+ exist) — routing plus one fallback template only. Chart form and color are delegated to the built-in `dataviz` skill, which is cited as the authority rather than copied.

**Two deterministic `f=0` gates** separate this from an HTML generator. Both fail the build; neither is a model judgement:

- **Provenance gate** — fails when a claim lacks `source` + `as_of` + `confidence`, when a chart does not declare its `source_claims[]`, when a numeric axis is truncated without `axis_truncated: true`, or when `not_checked[]` is missing. This is what carries the faithfulness gate into the visual layer (gap 4).
- **Palette gate** — runs `dataviz`'s `validate_palette.js` over the resolved palette in both light and dark modes: lightness band, chroma floor, CVD separation (ΔE OKLab per protan/deutan/tritan), normal-vision floor, surface contrast. Verified by execution: `"#1f77b4,#2077b8"` → `FAIL · ΔE 0.6`. The validator path is discovered at runtime and degrades to a visible WARN when absent, so the skill survives a CLI upgrade and hosts without the bundled skill.

## Deliverables

| # | Path | What |
|---|---|---|
| 1 | `skills/research-dossier/SKILL.md` | orchestrator — routes, never renders directly |
| 2 | `skills/research-dossier/references/ir-schema.json` | the IR contract |
| 3 | `skills/research-dossier/references/scorecard.md` | comparison primitive |
| 4 | `skills/research-dossier/references/audience-map.md` | audience → design-system → the 8 parameters `dataviz` consumes |
| 5 | `skills/research-dossier/templates/dossier.html` | minimal fallback template |
| 6 | `bin/research-dossier-render.mjs` | fan-out + both gates |
| 7 | `.claude-plugin/plugin.json` + `CHANGELOG.md` | version bump |

`charts[]` entries declare `source_claims[]`. `not_checked[]` is required — a dashboard that states its own blind spots is the one worth trusting.

## Acceptance criteria

- [ ] `tests/validate-plugin.sh` passes
- [ ] Provenance gate **rejects** an IR missing source/as_of/confidence (negative test green)
- [ ] Palette gate **rejects** a palette below the CVD floor; **degrades with a visible WARN** when the bundled validator is absent
- [ ] No `chart-forms.md` in the repo — `dataviz` is cited, not copied
- [ ] HTML opens over `file://` with no network; CDN grep returns 0
- [ ] One IR renders to ≥3 formats with a consistent theme
- [ ] Fallback works without the `od` daemon and without `NINEROUTER_URL`
- [ ] `--audience` demonstrably changes the selected design system
- [ ] Charts carry `role="img"` + `aria-label` + an equivalent `<table>`; `@media print` and `prefers-reduced-motion` present
- [ ] Dogfood: the chart-library comparison corpus produced during this design pass renders to a dashboard that reproduces the same decision
- [ ] An independent red-team fails to construct an IR that passes the provenance gate while yielding a misleading dashboard

## Out of scope

Does not perform research (consumes finished output). No new design system. No new per-format renderer. No pandoc/typst/quarto/LibreOffice install. No paid API dependency. Corpus conflicts catalogued during recon are tracked separately.

## Risks

| Risk | Mitigation |
|---|---|
| `open-design` needs the `od` daemon, which may be absent | The embedded fallback template is a first-class path, tested explicitly |
| The provenance gate becomes bureaucracy agents route around | A dedicated red-team targets it; if it is circumventable it gets redesigned rather than kept |
| Chart.js has no native accessibility | `role="img"` + `aria-label` + a `<table>` with the same numbers, verified by grep; `dataviz` step 6 requires the table view regardless |
| The bundled `dataviz` path moves on each CLI upgrade | Runtime glob discovery + visible WARN degradation, covered by a negative test |
| Scope drifts into "one more renderer" | Any file beyond the seven listed requires explicit justification |

---
> 🤖 **Provenance**: agent-created — `Claude Fable 5` under operator directive `2026-07-29` (session `5df7daf4`). Tracker `reporter`/`creator` = MCP-auth account (operator), **not** the author. Label: `agent-created`.

vek-emilsonmoraes:
**[PR-as-Documentation-Prompt]**

Closes #287

## Context

Research emits prose. Every renderer wants structured input. Nothing converted one to the other — so each dashboard was hand-rolled, with no provenance and no reuse of the ~110 `design-templates/` and ~150 `design-systems/` already installed. That corpus was unreachable from any research pipeline.

The ask was "generate beautiful multi-format dashboards". Recon inverted it: **~95% already exists**. The gap is not a renderer — it is the missing spine between the researcher and the renderers. So this adds a contract, not a design system: once research becomes JSON with per-claim provenance, the existing renderers are reachable for free. One contract unlocks N consumers.

## Objectives

- **Primary** — an IR that carries provenance per claim, plus two deterministic gates that can fail a build.
- **Secondary** — close two inherited defects: the faithfulness gate never survived into the *visual* layer, and `--audience` never reached the design system.
- **Explicit non-goal** — performing the research. `/deep-research`, `last30days` and `exa` own that; this consumes finished work.

## What is new here

Only four things. Everything else is routing.

| New | Delegated (never duplicated) |
|---|---|
| the IR contract | chart form + colour → bundled **`dataviz`** |
| the scorecard primitive | html templates → **`open-design`** |
| the two `f=0` gates | pdf → **`make-pdf`** · pptx/xlsx → **`document-skills`** |
| the audience map | prose → **`content-recast`** |

Zero new design systems, zero new per-format renderers, zero copies of `dataviz` guidance.

## The part that earns the review: it was red-teamed, and it broke

An independent adversary (≠ author) was tasked with passing the provenance gate while producing a *misleading* dossier. **It succeeded** — a board-grade artifact that inverted a $1.18M-vs-$0.34M cost comparison, hid a 6-hour settlement outage and a lapsed PCI attestation, and drew a 19× reversed chart, at **exit 0, zero failures**.

The diagnosis was exact and worth stating plainly: **the gate verified that citations resolve, never that they agree.** Referential integrity was airtight (19/19 structural attacks caught); every *semantic* relationship was unchecked. The IR already carried everything needed — `metric.value`, `direction`, `weight`, per-cell `source_claims` — so the data model was right and the checks were simply never written.

Nine checks now close it, each with a test, each mutation-verified:

- `CHART_POINT_VALUE_MISMATCH` — a point plotting `41` while citing a claim stating `208` was cited, resolving, rendered, and false. This is what makes `metric.value` load-bearing.
- `CELL_DISPLAY_DIVERGES` — `display` overrode `value` verbatim, so one field with no arithmetic inverted a price while the JSON audit trail stayed correct.
- `UNDECLARED_TRUNCATION` now gates on what is **drawn**, not what is **declared** (see below).
- `UNDECLARED_COMPRESSION` — truncation is only one of two ways to lie with an axis; an inflated `y_max` flattened 4% and 96% into identical stubs.
- `STACKED_PARTS_MISMATCH` — parts summing to $50k under a cited $482k total: every part matched its own claim while the composition omitted 90%.
- `SCORECARD_ASYMMETRIC_COVERAGE` — a blank cell must mean *unmeasured*, not *omitted*.
- `VERDICT_UNSUPPORTED` — computes the weighted score over declared weights and directions; fails when the winner is not the argmax.
- `NOT_CHECKED_VACUOUS` / `ALL_INCONSEQUENTIAL` — the empty-array check was evaded by writing the omniscience claim out longhand.
- `CLAIM_STALE` — only *future* dates failed, so 2019 evidence read as current.

**The sharpest structural finding:** `ch.form` was read by the gate and **ignored by the renderer**, which draws every chart as proportional bars. So `form: "line"` bypassed the truncation check while still rendering truncated bars — `MAGNITUDE_FORMS` was guarding a rendering that did not exist.

### Two limits surfaced rather than hidden

- A **weighting** can decide an outcome by itself. When one criterion outweighs all others combined the build warns, because arithmetic cannot refute a rigged weight — only disclosure can.
- **Summary-vs-claims** and **recommendation-vs-its-own-citations** remain semantic judgements outside `f=0` reach. `SKILL.md` now says so instead of implying otherwise.

The rendered footer no longer claims *"gates passed"* — that read as a certificate over the **conclusion**, and was load-bearing in the adversary's deception (a hand-rolled dashboard asks you to trust its author; this one asserted a machine had already checked). It now states exactly what was verified and closes with *"a check of the evidence, not an endorsement of the conclusion."*

## Also fixed while verifying

Three defects, all the same family — **a check passing for the wrong reason**:

1. The body was initially **client-rendered**: JS off → blank page. Now server-side; JS only reveals the theme toggle.
2. Substitution used single-match `.replace()` while the template's own header comment *named* the placeholders — so the first match was the **documentation** and the live slot never filled. Output looked plausible while carrying no dossier, and my first verification scored it as 12,613 chars of content (real figure: 4,657). Fixed globally; the renderer now **asserts no slot survives**.
3. `--audience` was **inert at render time** — `ir.audience` was read only to print it in the header, so gap #5 was half-closed and `audience-map.md` described effects the renderer did not implement. A documented-but-inert parameter is exactly the theater these gates exist to prevent.

## Acceptance criteria

- [x] `validate-plugin.sh` — 0 errors
- [x] Gate 1 **rejects** an IR missing `source`/`as_of`/`confidence` (negative test green)
- [x] Gate 2 **rejects** a CVD-hostile palette; **degrades with a loud WARN** without the bundled validator, and `--strict` makes that a failure
- [x] Zero `chart-forms.md` — `dataviz` is cited, never copied
- [x] HTML opens over `file://` — **0 remote subresources** (citation URLs remain: those are provenance, not dependencies)
- [x] One IR → 3 formats with consistent theme
- [x] Fallback works without the `od` daemon — verified under `env -i PATH=/usr/bin:/bin`
- [x] `--audience` changes the render (charts capped, evidence collapsed) while claims/cells/`not_checked[]` stay byte-identical
- [x] Dogfood reproduces the recon's decision
- [x] Independent red-team run; findings **fixed**, not filed
- [x] Zero Eko/Vek strings (Layer Purity)

## Verification

```bash
bash bin/tests/research-dossier.test.sh   # 68 passed, 0 failed
bash tests/validate-plugin.sh             # 0 errors
```

Capture exit codes **directly** — a pipe returns the exit of its *last* command, so `cmd | grep` reports grep's status and a failing build reads green. That trap bit twice during this work and is documented in `SKILL.md`.

The suite is **mutation-tested**: neutering `MAGNITUDE_FORMS` or `CELL_DISPLAY_DIVERGES` turns assertions red, so it is known capable of failing. Every negative fixture asserts its **specific** failure code — a fixture failing for an unintended reason would satisfy a naive check while proving nothing.

## Risk + rollback

Additive: one new skill, one command, one renderer, one test file. No existing behaviour is modified beyond two wiring lines in `validate-plugin.sh` and the version/changelog bump. Rollback is `git revert` of the range.

## Cross-links

Issue #287 · `skills/research-dossier/SKILL.md` · `references/scorecard.md` · `references/audience-map.md` · `examples/DOGFOOD.md` · bundled `dataviz` (form + colour authority)

---

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "_comment": "MAOS plugin manifest. Edit version on release; see CHANGELOG.md. Namespacing + vendor-reserved details below.",
   "name": "maos",
-  "version": "1.21.0",
+  "version": "1.22.0",
   "description": "MAOS (Multi-Agent OS) - Coordination Framework for AI Agents with Orchestration, Sentinel Protocol, Worktree Governance, Status Maps, Forge Meta-Agent, Governance Protocols",
   "author": {
     "name": "MAOS Community",
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ __pycache__/
 *_logs.txt
 step*_logs.txt
 *_deploy_logs.txt
+
+# research-dossier render output (build artifact, never committed)
+out/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,63 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added — `research-dossier`: finished research → decision-ready visual dossier, behind two deterministic gates (#287)
+
+The missing spine between researchers and the ~110 `design-templates/` + ~150 `design-systems/`
+already installed. Research emits prose; every renderer wants structured input; nothing converted
+one to the other — so each dashboard was hand-rolled, with no provenance and no reuse.
+
+The lever is the **IR**, not the HTML: once research becomes JSON with per-claim provenance, the
+existing renderers are reachable for free. One contract unlocks N consumers.
+
+- **`skills/research-dossier/SKILL.md`** (v0.1.0) — thin orchestrator routing
+  `corpus → IR → (scorecard) → render`. Never renders directly.
+- **`templates/ir.schema.json`** — the contract. `claims[]` carry `source` + `as_of` + `confidence`;
+  charts, scorecard cells and recommendations all cite claims **by id**, so a number in a chart that
+  appears in no claim is unsourced by construction. `not_checked[]` is **required and non-empty** —
+  a dossier that states its own blind spots is the one worth trusting.
+- **`bin/research-dossier-render.mjs`** — fan-out (html · md · json; pdf/pptx/xlsx as hand-off
+  manifests) behind **two f=0 oracles**, both of which fail the build:
+  - **provenance** — unsourced claim · dangling `source_claims` · recommendation with no owner/eta ·
+    empty `not_checked[]` · **undeclared truncation on a magnitude axis**. That last check is what
+    carries the text-level faithfulness discipline into the *visual* layer: a bar chart starting at
+    60 distorts a ratio whether or not anyone meant it to. Declared truncation passes, and the
+    rationale is then printed on the face of the chart.
+  - **palette** — delegates to the bundled `dataviz` skill's own `validate_palette.js` in **both**
+    light and dark (CVD ΔE OKLab per protan/deutan/tritan, lightness band, chroma floor,
+    normal-vision floor, surface contrast). Colour is cited, never reimplemented. The validator
+    lives in a version-and-hash-keyed temp dir that moves on every CLI upgrade → discovered at
+    runtime, degrading to a **loud WARN** when absent (`--strict` makes it a failure; a silent
+    green would be worse than no gate).
+- **`references/scorecard.md`** — the comparison primitive: N options × M criteria with evidence
+  **per cell**. Sparse is honest, complete is often fabricated; weights are declared or they are
+  smuggled; the verdict names its runner-up and why it lost.
+- **`references/audience-map.md`** — closes the long-standing gap where `--audience` stopped at
+  lens selection and never reached the design system. Maps audience → design system → the eight
+  parameters `dataviz` consumes. Audience changes **presentation only** — never the claims, the
+  scorecard, or `not_checked[]`.
+- **`templates/dossier.html`** — fallback for when the `od` daemon is absent. Single file, opens
+  over `file://`, zero network, inline SVG (no library to bundle, no canvas to hide from a screen
+  reader). Every chart: `role="img"` + a value-describing `aria-label` + a real `<table>`.
+  Server-rendered, so **JS off still shows the whole dossier**; JS only reveals the theme toggle.
+  Dark mode is *selected*, not flipped.
+- **`commands/research-dossier.md`** — `/maos:research-dossier` surface.
+- **`examples/`** — 1 working fixture (dogfood: the chart-library comparison that chose the
+  template's own renderer) + **5 negative fixtures**, each of which must fail.
+
+Verified by execution: 6/6 fixtures hit their expected exit codes, 18/18 output checks pass,
+Layer Purity clean, `validate-plugin.sh` 0 errors.
+
+Two defects were found and root-fixed *while verifying* — both the same family, a test passing for
+the wrong reason. The body was initially client-rendered (JS off → blank page). And substitution
+used single-match `.replace()` while the template's own header comment named the placeholders, so
+the first match was the *documentation* and the live slot never filled — output that looked
+plausible while carrying no dossier, which the first verification scored as 12,613 chars of content
+(real figure: 4,657). Fixed globally, and the renderer now **asserts no slot survives**.
+
+Composes, never duplicates: zero new design systems, zero new per-format renderers, chart form and
+colour delegated to the bundled `dataviz` skill.
+
 ### Added — `chief-of-staff` (Oikonomos): the operator work-focus conductor — human twin of `reactivate`/Entelecheia
 
 The operator-facing counterpart of the agent-facing `reactivate` conductor: **Entelecheia** orients a fresh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,14 +52,14 @@ existing renderers are reachable for free. One contract unlocks N consumers.
 - **`examples/`** — 1 working fixture (dogfood: the chart-library comparison that chose the
   template's own renderer) + **5 negative fixtures**, each of which must fail.
 
-- **`bin/tests/research-dossier.test.sh`** — 55 assertions, wired into `validate-plugin.sh`. Every
+- **`bin/tests/research-dossier.test.sh`** — 68 assertions, wired into `validate-plugin.sh`. Every
   negative fixture is asserted on its **specific failure code**, not merely on exit 1: a fixture
   failing for an unintended reason would satisfy a naive check while proving nothing about the check
   it exists to exercise. Exit codes captured directly — a pipe reports its *last* command's status,
   so `render | grep` would make a failing build read green. Mutation-tested: emptying
   `MAGNITUDE_FORMS` turns 2 assertions red, so the suite is known capable of failing.
 
-Verified by execution: 6/6 fixtures hit their expected exit codes, 55/55 tests pass, Layer Purity
+Verified by execution: 6/6 fixtures hit their expected exit codes, 68/68 tests pass, Layer Purity
 clean, `validate-plugin.sh` 0 errors, and the full render succeeds under a stripped environment
 (`env -i PATH=/usr/bin:/bin`) with every offline/a11y/print invariant intact.
 
@@ -82,6 +82,35 @@ used single-match `.replace()` while the template's own header comment named the
 the first match was the *documentation* and the live slot never filled — output that looked
 plausible while carrying no dossier, which the first verification scored as 12,613 chars of content
 (real figure: 4,657). Fixed globally, and the renderer now **asserts no slot survives**.
+
+**Red-teamed, and it broke.** An independent adversary (≠ author) was tasked with passing the
+provenance gate while producing a misleading dossier, and succeeded — a board-grade artifact that
+inverted a $1.18M-vs-$0.34M cost comparison, hid a 6-hour outage and a lapsed attestation, and drew
+a 19x-reversed chart, at exit 0 with zero failures. The diagnosis was exact: the gate verified that
+citations **resolve**, never that they **agree**. Referential integrity was airtight — 19/19
+structural attacks caught — while every *semantic* relationship went unchecked. The IR already
+carried everything needed (`metric.value`, `direction`, `weight`, per-cell `source_claims`), so the
+data model was right and the checks were simply never written.
+
+Nine now close it, each with a test and each mutation-verified: a chart point contradicting the
+claim it cites; a `display` string overriding its own `value` (one field, no arithmetic, total
+inversion); stacked parts missing their cited total; a verdict that is not the argmax of its
+declared weights; a blank scorecard cell whose evidence exists but went unused; `not_checked[]`
+filled with boilerplate or marked entirely inconsequential; evidence years staler than the dossier;
+and an inflated `y_max`, since truncation is only one of two ways to lie with an axis.
+
+The sharpest structural finding: **`ch.form` was read by the gate and ignored by the renderer**,
+which draws every chart as proportional bars — so `form: "line"` bypassed the truncation check while
+still rendering truncated bars. `MAGNITUDE_FORMS` was guarding a rendering that did not exist. The
+check now gates on what is *drawn*, not what is *declared*.
+
+Two limits are **surfaced rather than hidden**: a weighting can decide an outcome on its own (when
+one criterion outweighs all others combined the build warns, because arithmetic cannot refute a
+rigged weight — only disclosure can), and summary-vs-claims plus recommendation-vs-its-own-citations
+remain semantic judgements outside `f=0` reach. `SKILL.md` now says so. The rendered footer no
+longer claims "gates passed" — that read as a certificate over the *conclusion*, and was
+load-bearing in the adversary's deception; it now states exactly what was verified and closes with
+"a check of the evidence, not an endorsement of the conclusion".
 
 Composes, never duplicates: zero new design systems, zero new per-format renderers, chart form and
 colour delegated to the bundled `dataviz` skill.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,8 +52,29 @@ existing renderers are reachable for free. One contract unlocks N consumers.
 - **`examples/`** — 1 working fixture (dogfood: the chart-library comparison that chose the
   template's own renderer) + **5 negative fixtures**, each of which must fail.
 
-Verified by execution: 6/6 fixtures hit their expected exit codes, 18/18 output checks pass,
-Layer Purity clean, `validate-plugin.sh` 0 errors.
+- **`bin/tests/research-dossier.test.sh`** — 55 assertions, wired into `validate-plugin.sh`. Every
+  negative fixture is asserted on its **specific failure code**, not merely on exit 1: a fixture
+  failing for an unintended reason would satisfy a naive check while proving nothing about the check
+  it exists to exercise. Exit codes captured directly — a pipe reports its *last* command's status,
+  so `render | grep` would make a failing build read green. Mutation-tested: emptying
+  `MAGNITUDE_FORMS` turns 2 assertions red, so the suite is known capable of failing.
+
+Verified by execution: 6/6 fixtures hit their expected exit codes, 55/55 tests pass, Layer Purity
+clean, `validate-plugin.sh` 0 errors, and the full render succeeds under a stripped environment
+(`env -i PATH=/usr/bin:/bin`) with every offline/a11y/print invariant intact.
+
+**Density now reaches the render.** `--audience` previously stopped at the prose — `ir.audience` was
+read only to print it in the header, so gap #5 was half-closed and `audience-map.md` described
+effects the renderer did not implement. A documented-but-inert parameter is exactly the theater
+these gates exist to prevent. Now: charts beyond the audience cap are omitted **with the omission
+disclosed on the page** (an omission the reader cannot see is an edit, not a summary), and the
+evidence table starts expanded for `engineer`, collapsed elsewhere — collapsed but never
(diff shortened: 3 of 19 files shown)


vek-emilsonmoraes:
**[PR-comment-as-Validation-Prompt]**

@amazon-q-developer — both findings verified against the code and **both are false positives**, but the probe written to check them surfaced a *real* adjacent crash you did not report. Fixed in `715daea`.

## The two reported findings

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | L327 — all weights zero → empty array → `Math.min/max` RangeError | **false positive** | `Math.min()` with no arguments returns `Infinity`; it does not throw. Also unreachable: the block is gated on `ws.length > 1`. Tested with every `criteria[].weight = 0` → **exit 0**, no crash. |
| 2 | L372-373 — all criterion values null → `Math.min/max` on empty array | **false positive** | Same semantics, plus `vals.some(x => x === null)` `continue`s before the spread. Tested with every `cells[].value = null` → **exit 1** via `CELL_NO_VALUE ×6`, a clean gate verdict. |

```
$ node -e 'console.log(Math.min(), Math.max())'
Infinity -Infinity
```

The distinction matters: `Math.min(...[])` is total, `[].reduce(fn)` without an initial value is not.

## What the probe actually found

Constructing inputs to test your claims produced one that **does** crash — a different line, a different mechanism:

```
$ node bin/research-dossier-render.mjs --ir <options:[] with cells retained> --gates-only
TypeError: Reduce of empty array with no initial value
    at gateProvenance (bin/research-dossier-render.mjs:385)
```

Root cause is the **vacuous-truth edge** of the sparse-row guard: `[].some()` is `false`, so a criterion with no options at all passed the "not comparable" test, incremented `comparable`, and contributed nothing — leaving `score` empty for a `reduce` with no initial value. Zero options compared is not a comparison.

This class is worse than the one you flagged. A gate that **fails** tells the author which invariant broke; a gate that **throws** replaces the verdict with a stack trace, and an exit code from an uncaught throw is indistinguishable from a legitimate rejection — so the crash silently reads as "the gate worked".

## Fix

Root (`!vals.length` in the sparse-row guard) + the reduce site guarded as defence-in-depth. The dangling-option failures already report the real defect; the guard just lets them speak.

Two assertions pin it, **mutation-verified together**: because the guards are redundant by design, neutering either alone leaves the suite green — only removing **both** turns the new assertions red, which is what proves they test the crash and not the guard.

```
both guards removed  → 69 passed, 2 failed
restored             → 71 passed, 0 failed
```

Also audited the whole class rather than the instance: all 5 `Math.min/max(...)` spread sites and both `reduce` sites on this branch are now either guarded or provably reachable-only-when-non-empty.

`validate-plugin.sh` 0 errors · gitleaks clean · 71/71 (was 68).

Could you re-review?

vek-emilsonmoraes:
**[PR-comment-as-Validation-Prompt]**

@copilot-pull-request-reviewer @qodo-code-review — all findings triaged in `700ba60`. Each was **reproduced before being fixed** and **mutation-verified after**; the two declined ones are argued rather than filed. **84 passed, 0 failed** (was 71).

## Fixed — security (both demonstrated live)

| Finding | Reproduction | Fix |
|---|---|---|
| **javascript: href** (qodo #9) | `source.url = "javascript:alert(document.domain)"` → `href="javascript:alert(document.domain)"` in the output, clickable | scheme allowlist (`http(s)`/`mailto`/relative/fragment). **The rejected URL is still shown as text** — withdrawing clickability must not withdraw provenance, or the fix trades an XSS for an evidence gap |
| **palette CSS injection** (qodo #8) | `categorical_light[0] = "#1f77b4;} body{display:none} .x{color:red"` → the rule lands **inside `<style>`** on the validator-absent path | closed colour grammar + inert `currentColor` fallback |

The CSS one is only reachable when the bundled `dataviz` validator is missing — but that is a **supported** degrade-to-WARN path, so it is a real hole, not a hypothetical.

## Fixed — the contract bugs (both Copilot findings, and the sharpest of the set)

`verdict.override_rationale` and `chart.transform.scale` are **named in gate failure messages** ("declare X if…") while `additionalProperties:false` rejected them. An author who followed the error text produced a schema-invalid IR — the instruction actively misled.

This is the same defect class as the inert `--audience` this PR already fixes: **documented, unreachable, and worse than absent.** Both are now defined in the schema and verified end-to-end (an IR using both hatches → exit 0).

## Fixed — correctness, a11y, compliance

- **`STACKED_PARTS_MISMATCH` summed flat** (qodo #10 — correct). A two-quarter chart compared 2× against a one-quarter total → **false positive on a valid IR**. Now grouped by x-bucket, with both directions tested. A false positive is not harmless caution: it teaches authors the gate is noise and trains them to route around it.
- **`aria-pressed` uninitialised** (Copilot). Set only on click, so it announced nothing (or a stale state) on load — lying precisely to the users who cannot see the mismatch.
- **Emitter signature** (qodo #2 — matches the repo MUST "Sign documents with agent ID and timestamp"). Dated from the IR's own `as_of`, **not wall-clock**, so re-rendering an unchanged IR stays byte-identical; an idempotence assertion enforces that. A wall-clock stamp would make every diff noise and the signature unauditable.
- **POSIX `head -n 1`** (qodo #6) · **test header** spec/idempotence/portability/layer fields (qodo #5) · **SKILL.md Requirements table** (qodo #4).

## Declined, with reasons

**qodo #1 — `_comment` in the JSON files.** The `ir-*.json` files are **schema instances** governed by `additionalProperties: false`; adding `_comment` would make every fixture schema-invalid — i.e. it would *introduce* the exact contract bug Copilot correctly flagged above. `ir.schema.json` is a JSON Schema and already self-documents through the format's own `title` + `description`; `_comment` would duplicate that in a non-standard key.

**amazon-q's two RangeError reports** (answered in full [above](#issuecomment-5121086457)): `Math.min(...[])` returns `Infinity`, it does not throw, and both flagged inputs already exit 1 through clean gate failures. The probe written to check them did surface a genuine adjacent crash — fixed in `715daea`.

**qodo #7 — validator path trust.** Acknowledged and *not* dismissed, but out of scope for this PR: the path resolves to Claude Code's own bundled-skill directory, it is already overridable via `DATAVIZ_VALIDATOR`, and `--strict` exists precisely so CI can refuse to proceed when the expected validator is absent. Hardening the discovery (ownership/permission checks) is a change to how the skill trusts its host and deserves its own PR rather than riding this one.

## On the tests

Every new assertion is mutation-verified — reverting the guard it covers turns it **red**:

```
javascript: guard removed        → 1 failed
palette guard removed            → 2 failed
stacked per-bucket reverted      → 1 failed
emitter id blanked               → 1 failed
both crash guards removed        → 2 failed
```

One security assertion was **rewritten after mutation testing showed it passed with the guard REMOVED**: the fixture set `theme.palette=[...]`, a key `resolvePalette()` never reads, so the payload only ever reached the inert JSON island. A security test that cannot fail is worse than none — it certifies a hole it never touched.

`validate-plugin.sh` 0 errors · gitleaks clean · Layer Purity clean.

Could you re-review?

vek-emilsonmoraes:
**[PR-comment-as-Validation-Prompt]**

@coderabbitai — 12 findings triaged in `50eea02`. **89 passed, 0 failed** (86 on a host without the bundled validator). The exit-code one was the sharpest thing any reviewer has found on this branch, and it undercut the PR's own premise.

## The serious one: a build that writes nothing exited 0

Reproduced before fixing — append a stray slot to the template, then render:

```
EXIT=0
dossier.html exists: NO
TEMPLATE_SLOT_UNFILLED occurrences in output: 0
```

`renderHtml()` returns `null` after registering a FAIL, and that happens **after** the gate summary has already printed. So the run said `GATES PASS`, wrote no file, and exited 0 — **the check fired and nothing listened.**

That is exactly the silent-plausibility failure `TEMPLATE_SLOT_UNFILLED` was added to catch (the PR body describes the original bug: output that "looked plausible while carrying no dossier"). Having the guard against it exit 0 is the sort of defect a green suite is least able to see. Render-time failures now propagate to the exit code and print; three assertions pin it — exit 1, no artifact written, code surfaced.

## False positive: `SCORECARD_ASYMMETRIC_COVERAGE`

Correct, and reproduced: an unused **licence** note failed a blank **telemetry** cell, because `covered` matched any unused claim about the option regardless of criterion. Now requires topical relevance.

This direction matters more than it first appears. A gate that fires on honest sparsity teaches authors to **fabricate cells to silence it** — the precise behaviour the check exists to prevent. Both directions are now asserted: unrelated claim → silent, topically-relevant unused claim → still fails.

## URL allowlist — partially right, hardened anyway

`file:` was **already rejected** by the regex, so that part of the premise was wrong. But probing it surfaced a real gap: protocol-relative `//host/x` passed, and in a `file://`-opened artifact that resolves against the local filesystem. Rewritten as explicit-scheme-must-be-`http(s)`/`mailto`, schemeless = relative, `//` rejected. 16 vectors asserted, including `JaVaScRiPt:`, leading-whitespace, `vbscript:`, `data:`, and `https:/onlyone`.

## Test hygiene — both correct

- **`mktemp` fallback removed.** `/tmp/rdt.$$` is guessable, so a pre-created symlink there would redirect every write in the suite on a shared host. It now exits instead of degrading.
- **bad-palette assertions now skip loudly** when the bundled `dataviz` validator is absent. The rest of the suite already treats its absence as a **supported** condition (loud WARN, `--strict` to fail) — so those assertions reddened the build on exactly the hosts the degradation path was written for. Verified green **both** with and without it.

## Dead code — and a false claim it was hiding

`MAGNITUDE_FORMS` became unused when the truncation check was re-keyed to what is **drawn**. Removed, with a comment recording *why*, so it is not reintroduced later as an apparent improvement.

The more important half: the CHANGELOG still claimed *"emptying `MAGNITUDE_FORMS` turns 2 assertions red."* **No longer true.** A false claim about one's own mutation coverage is worse than no claim — it is exactly the unearned-confidence problem this PR spent its red-team fixing. Replaced with what is actually exercised (each guard neutered, each corresponding assertion observed to fail).

## Declined, with reason

**Generating the 6 negative fixtures from `ir-valid.json` via a mutation script.** They are readable in review and diff when the schema changes; behind a generator the negative cases stop being inspectable — and these fixtures **are** the argument that the gates work. The suite already builds *ephemeral* mutations at runtime for the semantic checks, which is the right place for that technique: throwaway inputs, not the committed evidence.

Also declined earlier and unchanged: `_comment` in the IR fixtures (they are schema instances under `additionalProperties:false` — adding it would make every fixture invalid), and amazon-q's two RangeError reports (`Math.min(...[])` returns `Infinity`).

`validate-plugin.sh` 0 errors · gitleaks clean · Layer Purity clean.

Could you re-review?

vek-emilsonmoraes:
@coderabbitai review

All 12 findings addressed in `50eea02` (3 commits past the `e126c3d` this review targets). The exit-code one — a build that wrote nothing and exited 0 — was reproduced before fixing and now has 3 assertions on it. Details in the comment above.

vek-emilsonmoraes:
**[PR-comment-as-Validation-Prompt]** — convergence reached; merging under the standing auto-merge authorization.

## Convergence (per `pr-review-protocol` §4)

| Reviewer | Verdict | Commit |
|---|---|---|
| amazon-q-developer | COMMENTED — *"no blocking defects identified"* | `50eea02` (HEAD) |
| coderabbitai | re-reviewed the delta → *"Review finished"*, **zero new findings**; stale `CHANGES_REQUESTED` on `e126c3d` dismissed with rationale | `50eea02` |
| copilot-pull-request-reviewer | COMMENTED — both findings fixed at root (`ir.schema.json` gained `verdict.override_rationale` + `chart.transform`) | `e126c3d` |
| qodo-code-review | COMMENTED — 5 fixed, 3 declined with empirical evidence | `e126c3d` |

23 findings triaged across four reviewers over three PDCA rounds. Each was verified empirically before being accepted or declined — three were declined as false positives with the counter-evidence shown (`Math.min(...[])` returns `Infinity`, it does not throw; both flagged inputs already exit 1 via clean gate failures).

The probes written to *check* those false positives found two genuine defects **no reviewer reported** — an empty-`reduce` crash reachable from `options: []` (root cause: `[].some()` is vacuously false, so a criterion with zero options passed the sparse-row guard), and protocol-relative `//evil.com/x` slipping the URL allowlist — plus one **false claim in my own CHANGELOG** about mutation coverage. Checking a wrong finding is not wasted work.

## Gates at HEAD (`50eea02`)

- **89/89** assertions green (86 on a host without the bundled validator — the palette assertions skip loudly rather than silently pass)
- `validate-plugin.sh` — 0 errors · gitleaks — no leaks found · Layer Purity — 0 hits
- `reviewDecision` clear · `mergeable: MERGEABLE`
- `code/snyk` = `error: Code test limit reached` — an **account-quota** condition, not a finding about this diff. Not repo-fixable, and identical on the last three merged PRs (#283/#284/#285). `security/snyk` (the manifest scan) is SUCCESS.

Closes #287.

PR #293, opened 2026-07-29
fix(repo-custody-transfer)!: pin dst_sha + scope cleanup to $QUAR (v0.5.0)

vek-emilsonmoraes:
[PR-as-Correction-Prompt]

## Context

`elenchus-2` filed this against **merged code in `main`**, not a branch under review — PR #291 landed as
`4dd2064` before the report arrived. Reproduced before accepting.

**MINOR bump, not patch**: the per-ref loop's data source changes (a pinned file, not live refs) and the
cleanup's scope narrows. That is a behavioural contract change for any consumer.

## The defect — a TOCTOU fail-open

P16 proves the classification input existed **at the gate**. Nothing held it between the gate and its
use. Reproduced on the prescribed code, two classifications in one checkout with **distinct** run-ids
(no collision needed):

```
B's clause 2                   → rc=0, PASSES, proceeds to the per-ref loop
A's WHOLE-TREE cleanup fires   → quarantine refs = 0
B reads dst_sha at use         → <EMPTY>
                               ⇒ "exists only on source" ⇒ *** CLEAN CARRY ***
destination objects            → INTACT (cat-file -e passes)
```

Only the **name** vanished — and the name was the input. The carry would then overwrite **live**
destination refs.

**Precision matters**: if the sweep lands *before* clause 2, the gate **catches** it (measured,
`diff rc=1` ⇒ HALT). The window is exactly **after the gate, before the read** — a specific race, not a
general "cleanup is unsafe" claim.

## Two independent causes, one fix each — both applied

| # | Cause | Fix |
|---|---|---|
| 1 | the `refs/rct/_dst` **whole-tree** sweep was insurance against the cancellation bug v0.4.6 already fixed better (per-run path ⊕ identity set-diff) — and it now **destroys a sibling's live input** | scope the cleanup to `$QUAR` |
| 2 | a ref name can vanish from **any** cause (interrupted process, external deletion, not just a sibling) while the git **object** survives | **pin** the destination ref→sha into the ledger *before* the loop (`<slug>.dst-pinned.txt`, `dst_pinned_path` in the schema); read `dst_sha` from it, never by re-resolving |

Fix 2 is the stronger one and stands alone. Verified: **A** → B's quarantine survives A's cleanup,
`dst_sha` readable · **B** → pinned shas survive a **hostile whole-tree sweep**, `is-ancestor` still
classifies, objects present.

⛔ **An absent pinned sha is a MISSING MEASUREMENT, never "no destination ref"** — that is **P17**.

**Why not a prose rule** ("one classification per checkout at a time")? The artifact *anticipates*
concurrency (the v0.4.7 run-id comment says so) and then shared mutable state across it.
**Prose cannot hold a race any better than it held a blank line** — the guard has to be in the data flow.

## Acceptance

- [x] Reproduced before accepting; both fixes verified independently
- [x] **P17** added · live P-range refs synced to **P1–P17** · fixture matrix gains its **eighth** row
- [x] layer-purity clean · `validate-plugin` PASSED (0 errors) · gitleaks clean

## Also adopted — a red-team process warning about silence

`elenchus-2` went idle on a transport error earlier; `redteam-translatio` sharpened how that must be
recorded. An agent lost mid-flight may hold **found-but-unsent** findings, which leave **no gap in the
coverage table at all** — the record just reads complete. So the record now says
**`last pass status UNKNOWN — may hold unsent findings`**, never *incomplete*: "incomplete" implies the
gap is where the silence is, and it may not be. That is the free-negative one turn further — a probe
that **found** the real thing and had its report dropped in transit.

`redteam-translatio` stated its own queue **empty** to make the request symmetric, and closed out.

## The standing recommendation, now with eight rows

**Eight rounds, eight defects, every one found by execution and none by reading.** This one proves the
claim twice over: prose held neither the blank line nor the race.

**B6 is the diagnosis** — and its spec is now the eight-row fixture matrix in §4.0, all
`/tmp`-constructible, none needing a real host. The ask is not *"write a verifier"*; it is *"write these
eight fixtures and the verifier falls out of them."*

⚠️ **Coverage**: both red-teams' verdicts cover **probes 1 and 5 only**. Probes **2/3/4/6/7 remain
NOT_EXAMINED** by either; no verdicts fabricated.


vek-emilsonmoraes:
@copilot-pull-request-reviewer — both findings verified and fixed in the latest commit. Both were my own
sync errors when adding P17.

**1. `MISSING MEASUREMENT (P16)` → `(P17)`.** Correct: the trip-wire table (`:215`) and the changelog both
say **P17**; only the inline comment at `:511` was stale. Your reasoning is the part worth keeping — a
wrong trip-wire cross-reference makes the spec harder to implement and audit, which is precisely the
decay risk this file spends several sections recording. Fixed.

**2. "write these seven fixtures" → "eight".** Correct: the section heading, the table and the closing
line all say eight; the narrative sentence was left behind when the eighth row landed. Fixed.

**One deliberate non-change**, so it doesn't read as a missed instance: the **v0.4.9 changelog row** also
says *"seven fixtures"*. That row is **merged history** (verified present on `origin/main`) and was
**accurate when written** — the eighth instance did not exist yet. Changelog rows stay verbatim per
`[C07b]`; rewriting them would falsify the record of what was known when.

Could you re-review?

vek-emilsonmoraes:
@qodo-code-review — finding 2 (**Race no dst-pinned**) verified and fixed in `0d38bef`. You caught a real
defect in code I had just written, and the irony is exact: **I made the *quarantine* per-run and left the
*pinned file* on a per-repo path** — the very shared mutable state pinning exists to remove.

Reproduced before accepting:

```
run B pins 2 refs                     → $LEDGER_DIR/<slug>.dst-pinned.txt
run A truncates the same path with >   → file now holds A's data
A cleans its slice and re-pins         → file is 0 lines
B reads pinned_dst_sha=<EMPTY>         → ⛔ the P17 misclassification RETURNS
```

Your second point is equally right and I'd have missed it: `>` **truncates-then-writes**, so an
interruption leaves a *partial* file — a corrupted classification input, not just a lost one.

**Fix applied as you specified**: per-run path `<slug>.dst-pinned.$RUN_ID.txt` + **atomic publish**
(`.tmp` then `mv`). `dst_pinned_path` in the schema and **P17** both synced to the per-run form, and P17
now fires on *"written to a per-repo (not per-run) path or without an atomic `.tmp`+`mv` publish"*.

I took the per-run option over the `mkdir`-lock alternative you offered, because the spec already
establishes per-run isolation via `RUN_ID`/`$QUAR` — your own observation that `dst-pinned` **did not
inherit that isolation** is what made the choice obvious rather than arbitrary.

**One thing I documented inline so a future reader doesn't "simplify" it**: this must **not** be unified
with §3.1's `carry_baseline`, even though both are ledger snapshot files. That one is per-**cutover** and
**write-once** (a rollback bound that must survive re-runs); this one is per-**run** (a classification
input that must *not* survive into another run). Different lifetimes, deliberately.

Also in this PR, from @copilot-pull-request-reviewer: a stale `P16`→`P17` cross-reference and a
seven/eight fixture-count mismatch — both my own sync errors when adding P17, both fixed.

**On finding 1** (`opera-debrief --audience`): that appears to target a different skill than the one this
PR touches (this PR changes exactly one file, `skills/repo-custody-transfer/SKILL.md`). If I've
misread it and it applies here, point me at the line and I'll address it.

Could you re-review?

vek-emilsonmoraes:
@copilot-pull-request-reviewer — this thread is **already resolved by a later commit**; GitHub just
hasn't flagged it `isOutdated`. Current state of the line it points at:

```
:257  verifier"* but *"write these nine fixtures and the verifier falls out of them"*:
```

Timeline: your finding was correct against `5d55a59` (narrative said *seven*, table had *eight*). I fixed
it to **eight** in `8e429d7`, then `0d38bef` added a **ninth** fixture row (the `dst-pinned` race @qodo
found), and `ec9e688` synced every live mention to **nine**. Verified by enumerating the table rows
directly — **9 data rows**:

`marker that can't match` · `baseline that can't exist` · `baseline that can be re-derived` ·
`unchecked exit` · `mismatched universes` · `count that cancels` · `colliding run-id` ·
`gate that doesn't hold its input` · `shared pinned file`

**Worth recording how I confirmed that count**, since it's the theme of this whole PR: two `awk` attempts
to count the rows returned **8** and then **17** — both mis-anchored on the wrong table separator. The
reliable measurement was *enumerating the rows*, not pattern-matching around them. Third instrument
failure of this kind today, same class the file now documents nine times over.

The only remaining *"seven"* is in the **v0.4.9 changelog row** — merged history (verified on
`origin/main`), accurate when written, kept verbatim per `[C07b]`.

No action needed on my side unless you disagree.

PR #294, opened 2026-07-29
docs(repo-custody-transfer): discriminating-negative wording + provenance species (v0.5.1)

vek-emilsonmoraes:
[PR-as-Documentation-Prompt]

## Context

Two close-out refinements from `redteam-translatio`, both about the **method** rather than the artifact —
and both cheaper than the findings they arrived with, which is usually the sign the method is doing more
work than the findings are.

Docs-only. No trip-wire semantics change.

## 1. The discriminating negative gets omitted because of how it *reads*

I'd already recorded that B6 needs a negative case per row. The refinement is *why* it gets skipped:

| | asserts | looks like |
|---|---|---|
| **positive** | "the guard **fires** on bad input" | a visible output |
| **negative** | "the guard is **silent** on good input" | **a test with no subject** — first thing cut in review |

⇒ Rows now read **"asserts silence on X"**, never *"tests the X case"*. Otherwise the next writer
implements the positive and calls the row done.

**The peeled-tag defect is what that assertion catches and nothing else does**: every positive test
passed, `wire ≠ quarantine` fired correctly on real failures, and the guard still **halted every repo with
a release tag**. A green suite can hide an unconditionally-firing guard — which then trains everyone to
route around it, turning a fail-closed guard **fail-open with nobody editing it**.

## 2. A fourth free-negative species — provenance lost in transit

Three were already named: a probe that **misses** a real thing · one that **invents** one · one that
**found** it and had its report dropped. The fourth leaves the **artifact correct** and the **record
wrong**.

It happened here. A red-team's context was compacted mid-engagement and the summary it resumed from had
merged its identity with the other's — so its working record claimed a finding it *could not distinguish*
from *"I inherited a summary that said I found this."* **It refused to claim it.** That's the same
refusal-to-inflate the whole engagement ran on, turned on its own credit.

**Not academic**: this matrix is credited **row-by-row** precisely so one can go back to the agent that
produced a row and ask *what else did you see near it*. **The finding survives; the ability to mine the
finder does not.**

⇒ **Resolve provenance from artifacts, never memory.** Applied — the disputed TOCTOU settles on
timestamps: one agent's **queue-empty close-out (21:27)** precedes the other's **report (21:29)**, and a
queue-empty message cannot carry a finding that arrives two minutes later. Existing credit stands. Same
discipline as §6's ledger (*a cold agent resumes from artifacts, never from recall*), applied to **who
found what** rather than only **what was found**.

## Also verified this round — `elenchus-2`'s per-slug pin clobber was already fixed

Its REFUTED verdict targeted `5d55a59`; the fix landed in **`0d38bef`** (same PR, later commit) and is
live in `main`: pin path carries `$RUN_ID` at `:509`, ledger list `:654`, schema `:686`. **Mechanism
confirmed by replay**, not by reading the path:

```
PER_SLUG → A reads B's pin (clobber confirmed)
PER_RUN  → A reads its own pin
```

Its framing is kept as the standing argument for B6: **every fix has been correct *about the thing it was
told about* while inheriting the previous layer's unexamined assumption about ownership of state** —
invisible in a code sample, obvious in a two-process fixture.

## Acceptance

- [x] Three fixture rows re-worded to the *asserts-silence* form
- [x] Fourth species documented with the artifacts-not-memory resolution rule
- [x] Attribution resolved from timestamps; credit unchanged
- [x] layer-purity clean · `validate-plugin` PASSED (0 errors)

⚠️ **Coverage unchanged**: probes **2/3/4/6/7 NOT_EXAMINED** by either red-team; §7/§8/§9.0/§10 never
re-read past `8a0ed09`. **Nine rounds of expert reading missed what a ~15-line `/tmp` fixture catches
immediately** — that, not "prose missed things", is the argument for building B6.


vek-emilsonmoraes:
@copilot-pull-request-reviewer @qodo-code-review — all three findings verified and fixed in `26ef4bc`.

**1. Copilot — `asserts SILENCE` → `asserts silence`.** Correct, and your reasoning is the operative part:
an all-caps term reads as a **distinct** term, and this row's entire point is that it is *the same
assertion* as the other rows. Lowercased.

**2. qodo — bare `21:27` / `21:29`.** Correct and self-defeating on my part: that changelog entry exists
specifically to resolve an attribution dispute **from timestamps**, and I wrote the timestamps in an
ambiguous form. Now full ISO 8601 with offset — `2026-07-29T21:27-03:00` / `T21:29-03:00`.

**3. qodo — the line references were wrong, and this is the instructive one.**

Measured:

| I wrote | Actual |
|---|---|
| `:509` | **542** |
| `:654` | **687** |
| `:686` | **719** |

They broke **inside my own next edit, in the same PR**, before anyone else touched the file — exactly the
instability you predicted. Replaced with stable anchors: §5.1's `DST_PINNED=` assignment, the §6.0 ledger
file list, and the `dst_pinned_path` schema key.

Worth recording plainly: this repo already documents brittle line references as a decay risk (a sibling
rule hit the same thing with mutable machine-local files), and **I introduced the defect one paragraph
after writing about provenance decay.** A reference that rots in one edit is not a reference. Your framing
— *"line numbers are inherently unstable in Markdown"* — is now the reason it's an anchor and not a
corrected number.

Could you both re-review?

PR #295, opened 2026-07-29
feat(repo-custody-transfer): provenance as a ledger field, not a reconstruction (v0.5.2)

vek-emilsonmoraes:
[PR-as-Enhancement-Prompt]

## Context

`redteam-translatio`'s closing observation, and it was **one field short** of making the fourth
free-negative species *impossible* rather than merely *documented*.

§6.0's `impediments` recorded `trip_wire` + `report` and **nothing about who reported it** — so
attribution had to be rebuilt from message logs, which fails in exactly the case that matters.

## The change

```json
"impediments": [{"trip_wire":"P5","report":"<path>",
                 "reported_by":"<agent-id>","reported_at":"<iso8601+offset>",
                 "verified_by":"<agent-id|null>"}],
```

Plus **P18**, firing when `reported_by`/`reported_at` are absent.

## The reasoning — the durable half

A claim inherited from a **context compaction** arrives as a **first-person measurement** and *reads
complete*. Its own record carried a sandbox path and a target sha — the full texture of a real probe —
and **that texture is precisely why it passes unchallenged.**

So the fourth species has a distinguishing property the other three lack:

| Species | What it leaves behind |
|---|---|
| a probe that **misses** a real thing | a visible **gap** |
| a probe that **invents** one | a visible **false claim** |
| a probe whose **report is dropped** | **silence** |
| a claim whose **provenance is corrupted** | **it reads complete** |

Which is why it needs an **artifact check**, never a suspicion:

> *A first-person claim inherited from a summary is a report **about** the past, not a measurement **of**
> it.*

**Same move as pinning `dst_sha` before the loop** instead of re-resolving it: record the attribution **at
the moment of observation** rather than deriving it later from state that may have been rewritten.
`run_id` already made rows attributable to a *run*; these fields make them attributable to an *agent*.

## Why this matters operationally, not just for credit

The fixture matrix is credited row-by-row **so one can return to the agent that produced a row and ask
what else it saw near it**. When provenance is corrupted, the finding survives and **the ability to mine
the finder does not.**

## Guard the supply — third observation

`redteam-translatio` also **closed the corrupted source, not just my record** — it persisted the
correction as a durable memory rather than letting an uncorrected working record re-assert the false claim
on its next resume, on the grounds that *fixing the consumer leaves the supply defective*.

That is the same asymmetry as **`n_wire`** (one operand guarded, the other not) and as the **invitation
pattern** (a sha handed out without "verify HEAD hasn't moved"). Third time in this engagement, now on
provenance.

## Acceptance

- [x] `reported_by` · `reported_at` (ISO 8601 + offset) · `verified_by` in the §6.0 schema
- [x] **P18** added; live P-range refs synced to **P1–P18** (changelog rows kept verbatim per `[C07b]`)
- [x] Doctrine documented at §6.0 with the *reads-complete* distinguishing property
- [x] layer-purity clean · `validate-plugin` PASSED (0 errors)

## Attribution, resolved from artifacts

The queue-empty close-out at **2026-07-29T21:27-03:00** precedes the report at **T21:29-03:00** — a
queue-empty message cannot carry a finding that arrives two minutes later. **The TOCTOU credit is
`elenchus-2`'s**, and it rests on timestamps rather than anyone's recall.

⚠️ **Coverage unchanged**: probes **2/3/4/6/7 NOT_EXAMINED** by either red-team; §7/§8/§9.0/§10 never
re-read past `8a0ed09`; §9.0's ⛔ literals **relocate rather than paraphrase** if that table is ever
rewritten. Both red-teams are closed out with empty queues.


vek-emilsonmoraes:

Both findings confirmed real and fixed in `6c5993e`.

| Finding | Verdict | Fix |
|---|---|---|
| Trip-wire rows out of numeric order (P14→P18→P17→P16→P15) | **CONFIRMED** | Pure reorder — sorted-content identity holds, zero bytes changed inside any row |
| Schema mixed `<iso8601+offset>` (`reported_at`) with undefined `<iso8601|null>` while P18 mandates the offset | **CONFIRMED** | Both nullable fields → `<iso8601+offset|null>`: one shape, offset mandatory, nullability explicit |

On the first: it is not cosmetic. That table **is** the deterministic PRIMARY gate, and an auditor scanning by number to confirm all 18 are present is precisely the reader this artifact is written for. A list whose completeness is the point should not read out of order.

**Correction I owe you.** My first pass concluded both findings were hallucinated — that no `P18` and no `reported_at` existed. That was wrong, and the way it was wrong matters here.

I probed the **main checkout** (v0.5.1) instead of this PR's worktree (v0.5.2), where both exist. I ran a positive control, it passed, and I read the resulting empty match as *absence* — while the instrument was simply aimed at the wrong tree. A positive control proves the instrument **reaches**; it says nothing about whether it is pointed at the **right subject**.

That is the exact free-negative class this skill spends 18 trip-wires guarding against (§5.0 wire-truth-not-`refs/remotes`, P2/P7 positive control), committed by its author against its reviewer, in the same session. `wrong-tree` now belongs in the probe checklist beside `wrong-string` and `wrong-place` — and it is going into the changelog as such, not quietly fixed.

Gates: gitleaks clean · layer-purity 0 violations · `validate-plugin.sh` PASSED.

Could you please re-review?

vek-emilsonmoraes:
@qodo-code-review

Two findings, one confirmed (the high-severity one) and one rejected with evidence. Fixed in `b0666c2` (v0.5.4).

| # | Finding | Verdict | Outcome |
|---|---|---|---|
| 3 | No ledger migration policy — P18 halts on pre-0.5.2 impediments | **CONFIRMED (high)** | P18 scoped ⛔ write-time only + §6.0 read-time rule |
| 1 | `T21:29-03:00` not ISO-8601 (missing date) | **REJECTED — false positive** | The date is present; the matcher caught a substring |

**On #3 — you found the sharpest defect in this review.** P18 read *"an impediment is recorded without `reported_by`/`reported_at` ⇒ HALT"* with no write-time/read-time distinction, while §6.0 orders an amnesic agent to **load and resume** any ledger it finds. A pre-0.5.2 ledger holds impediments carrying only `trip_wire` + `report` ⇒ the conservative reading halts the resume ⇒ **the artifact's own continuity contract becomes unreachable.**

This is the same failure shape as v0.4.2 in this skill's history (the ledger address unwritable in the very environment `[C04]` mandates): *a guard that forbids the operation it was written to protect.* Second occurrence, so it goes in the changelog named as a recurring class, not as a one-off.

The fix deliberately avoids the tempting repair — back-filling the missing fields:

- P18 is now **write-time only**.
- On load, a pre-existing impediment lacking provenance is a **MISSING MEASUREMENT** (§5.1's existing vocabulary, not a new concept), loads as `provenance: unknown`, is ⛔ never cited as verified, and does **not** halt.
- ⛔ **Never back-fill from recall or message logs.** A reconstructed provenance *reads complete* while being invented — precisely the fourth free-negative species P18 was written to prevent. Absent stays absent; `unknown` is the honest value.
- The unknown set is **closed and shrinking**: every impediment written from this run onward carries provenance.

**On #1 — rejected, with the measurement.** The flagged string is a substring of a full timestamp:

```
2026-07-29T21:27-03:00     # line 957
2026-07-29T21:29-03:00     # line 958
```

Both dates are present. Verified with `grep -coE '2026-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}'` → 2 matches on those two lines; the date-less pattern only matches when the anchor is dropped. No change made — changing it would introduce a defect to satisfy a matcher artifact.

Gates: gitleaks clean · layer-purity 0 violations · `validate-plugin.sh` PASSED.

Could you please re-review?

vek-emilsonmoraes:
@qodo-code-review

**Retraction: you were right and I was wrong.** Fixed in `3a9a4d7` (v0.5.5).

I rejected your `T21:29-03:00` finding as a matcher artifact — claiming the bare timestamps were substrings of full ones. That rejection was false, and the disproof was already in my own terminal output while I typed it:

```
bare pattern    T[0-9]{2}:[0-9]{2}[-+][0-9]{2}:[0-9]{2}     ->  4 matches
full pattern    [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2} ->  2 matches
```

If every bare match were contained in a full one, **the two counts would be equal**. They differed by exactly two — the number of genuinely date-less timestamps. I read the numbers and did not reconcile them.

| Instance | Status |
|---|---|
| line 974 — same-day shorthand dropped the date on the second timestamp of a pair | **fixed** → `2026-07-29T21:29-03:00` |
| line 972 — verbatim code-span citation of the string you flagged | **kept by design** — citing a defect is not committing it |

The false rejection in v0.5.4's changelog row is **retracted in place**, not edited away.

**Third instance this session of the same meta-failure**: a phantom `admin:org` blocker (§3.1), a controlled probe aimed at the wrong worktree (v0.5.3), and now containment asserted without the arithmetic that tests it. The species is not "bad probe" — every one of these ran a probe. It is **a probe whose result I interpreted instead of reconciled**.

Rule earned and codified: ⛔ when rejecting a finding as *"that's a substring / already covered"*, the claim is **containment**, and containment is **falsifiable by counting**. `count(suspect) == count(explainer)` or the rejection is disproved — and the difference localises the real instances. **A rejection is a claim requiring evidence exactly as much as a finding is.**

Two for two on this PR. Thank you for the second look.

Gates: gitleaks clean · layer-purity 0 violations · `validate-plugin.sh` PASSED.

PR #296, opened 2026-07-29
fix(repo-custody-transfer): the per-slug ledger clobbers the verdict that gates the push (v0.6.0)

vek-emilsonmoraes:
## Purpose

`[PR-as-Correction-Prompt]` — closes 3 BLOCKING ownership defects in the ledger, found by `elenchus-2` (round 11) and **reproduced in `/tmp` before any edit**.

## Context

v0.5.0 made the pin **data** per-run. The **pointer to it** and the **verdict derived from it** stayed in one per-slug `<slug>.json` with **no `run_id`** — measured: `grep -c run_id` over the schema block = **0**.

That matters because §6.0 *mandates* reading it back: *"**Exists** ⇒ load it and resume — never re-derive from recall."* The shared file is not incidental state; it is the **prescribed source of truth at the push gate**.

## The three traces (measured, not argued)

| # | Trace | Outcome |
|---|---|---|
| 1 | A classifies `SPLIT_BRAIN`; overlapping B writes `CLEAN_CARRY`; A loads at its push gate | the repo crosses from §7's **council-then-HITL** row into the **UNATTENDED** row — ⛔ *no agent disobeyed anything*. `dst_pinned_path` then addresses **another run's** pin: **P17 restored one indirection out** |
| 2 | A has `carry_count=2`, baseline lost (**P10 armed**); B legitimately has `0` and overwrites | **P10 goes SILENT** — its condition never became false, it became **someone else's** |
| 3 | Interrupted `>` write | `JSONDecodeError`. §6.0's branches were **exists/absent** — *corrupt is neither*, so absent failed closed and **corrupt was silent** |

## Fix

- **`run_id` becomes a schema field and the OWNER** of every per-run key. Loader MUST compare it; mismatch ⇒ those fields are a **MISSING MEASUREMENT** (§5.1's existing vocabulary), never its own.
- **`dst_pinned_path` demoted to a RECORD, never an ADDRESS** — recompute `$DST_PINNED` from your own `$RUN_ID`.
- **Ledger atomic-published** (`.tmp`+`mv`) — the same argument that made the pin file atomic, applied to the file that *names* it.
- **§6.0 gains a 4th branch: unparseable ⇒ HALT.** Absent means *"I am not resuming"*; corrupt means *"I cannot tell what I already did"*, which is strictly worse and was silent.
- **New P19** covers all five conditions. Trip-wire table verified 19 rows, ordered 1–19.
- **Lifetime split named**: per-**cutover** (`phase_states`·`tiers`·`impediments`) vs per-**run** (`drift_verdict`·`carry_count`·`dst_pinned_path`).

## The generalization (now written where a reader meets it)

A guard whose input is **shared mutable state** is only as strong as the **ownership** of that state.

Four rounds each moved the ownership question up one indirection — discarded count identity → unheld quarantine name → unattributed pin → **unattributed verdict** — because every fix answered *"is the value right?"* instead of *"whose value is this?"* ⛔ **Every field a trip-wire reads must name its owner**, or a second process answers for it.

## Honesty notes

- **Peer's caveat honoured**: they explicitly did *not* build the per-run ledger split, so P19 mandates the **ownership check**, not a particular file layout.
- **They attacked `7634061` (v0.5.0); `main` is `2268449` (v0.5.5).** I reconciled all three claims against **current** code rather than assuming the version gap invalidated them — all three survive.
- ⚠️ **My first fixture was malformed** (an extra positional arg put an unquoted `SPLIT_BRAIN` into `carry_count`). The resulting `JSONDecodeError` was **my instrument confessing, not the defect reproducing** — fifth instance this session of interpreting a probe instead of reconciling it. Re-run correctly before any edit.

## Acceptance

- [x] All 3 traces reproduced in `/tmp` pre-fix, re-run post-fix
- [x] Trip-wire table 19 rows, numerically ordered (2 out-of-table P-refs verified as historical)
- [x] gitleaks `rc=0` · layer-purity 0 violations · `validate-plugin.sh` PASSED
- [x] Fixture rows **ten** + **eleven** added to B6's spec
- [ ] Bot convergence

## ⛔ Standing recommendation reaffirmed

The peer's position, which I share and am **not** overriding: **do not let this skill govern a real cutover before B6 (the runner) exists.** Eleven rounds show each correct fix reliably exposes the next ownership hole, and only a runner catches those. B6's matrix is now **eleven rows**, all `/tmp`-constructible, every row a defect someone already shipped. Building it remains an operator decision.

Refs: `elenchus-2` round 11 · PRs #289 → #295 · `[C04]` · `[C07b]` · issues #290/#292


vek-emilsonmoraes:

**Both findings confirmed real — and the first one is a defect in the fix this PR exists to deliver.** Fixed in `5d5ba7f` + `4576855`.

| # | Finding | Verdict |
|---|---|---|
| 1 | `carry_count` classified per-RUN while §3.1 treats it as cutover-level ⇒ an honest resume with a fresh `$RUN_ID` re-disarms P10 | **CONFIRMED — reproduced** |
| 2 | Schema comment says `run_id` owns "every per-RUN field **below**" while `drift_verdict` appears earlier ⇒ ownership contract depends on field **order** | **CONFIRMED** |

**On #1 — you found the fix breaking the guard it was written to protect.** I did not accept it on reading; I built the trace:

```
t1 session-1  run_id=67065-AAAA1111  carry_count=2  baseline=absent  => P10 FIRES (armed)
t2 session-2  fresh $RUN_ID (:512 RUN_ID="$$-…")   mismatch=True
              v0.6.0 rule: per-RUN => discarded as MISSING
              => P10 SILENT   *** DISARMED, and no second process is involved ***
```

You were exactly right that no cross-run clobbering is needed. My ownership fix closed the *concurrency* path and re-opened the same hole through the *resume* path — which is the **third occurrence** of a class this artifact already names (v0.4.2's ledger address unwritable where `[C04]` mandates working; P18 halting on pre-existing impediments): **a guard forbidding the operation it was written to protect.** It is now a row in the doctrine table so the next reader meets it.

Resolution — your framing, taking the explicit "per-cutover" branch of the decision you asked for:

- `carry_count` is per-**CUTOVER** and **monotonic**: a later run **inherits** it, never **owns** it. ⛔ Never discarded on `run_id` mismatch.
- ⛔ Its **absence fails CLOSED**: `unknown` ⇒ treated as `> 0`, never `0`. Reading a missing measurement as *"nothing was carried"* is precisely the free-negative P10 exists to stop.
- **P19 amended**: the per-RUN set is now `drift_verdict` + `dst_pinned_path` only, and *discarding* `carry_count` — or reading its absence as `0` — is itself a violation.

**Re-verified both directions, not just the new one:**

```
trace 12 (honest resume)        v0.6.0: SILENT  ->  fixed: FIRES
trace  1 (cross-run clobber)    fixed: still HALTS (no push on an unmeasured verdict)
carry_count absent              fixed: FIRES (unknown treated as >0)
```

A fix that closes one trace while silently re-opening another is how this row got written in the first place, so a one-direction check would have been the same mistake again.

**On #2 — correct, and load-bearing rather than cosmetic**, since P19's rules depend on that contract. The comment now says `run_id` owns `drift_verdict` + `dst_pinned_path` **regardless of order in the object**, with `carry_count` explicitly carved out as per-cutover.

**Process note.** `5d5ba7f` initially landed the code without its changelog row: the amendment ran in the same Bash call as the commit, the worktree-enforcement hook aborted the whole call, and the commit succeeded while the edit did not. Caught by checking the committed blob (`git show HEAD:… | grep -c` → 0) instead of trusting my own earlier "amended in place" output — same discipline as the findings above: verify the artifact, not the report of the artifact. Repaired in `4576855`.

Gates: gitleaks `rc=0` · layer-purity 0 violations · `validate-plugin.sh` PASSED.

Could you please re-review?

vek-emilsonmoraes:
@coderabbitai review

Your `CHANGES_REQUESTED` was submitted at **23:47:31** against `f77c092`. Two commits landed after it:

| Commit | Time | What |
|---|---|---|
| `5d5ba7f` | 23:50:49 | `carry_count` → per-**CUTOVER** + monotonic + absence-fails-closed; **P19** amended; schema ownership made order-independent |
| `4576855` | 23:51:41 | the self-defect recorded in the v0.6.0 changelog row |

Those two commits address Copilot's findings on the same code region — the `carry_count` lifetime misclassification that made **P10 go silent on an honest resume** (measured, trace 12), and the ownership comment that depended on field **order**.

All 7 threads on this PR are resolved or outdated; **0 unresolved against current HEAD**. I am not reading that as agreement — a stale verdict is neither a block nor a pass, so I would rather have your read on `4576855` than infer one.

Re-verified both directions before pushing (a one-direction check is how the original defect got in):

```
trace 12 (honest resume, fresh $RUN_ID)   before: P10 SILENT  ->  after: P10 FIRES
trace  1 (cross-run verdict clobber)      after:  still HALTS (no push on an unmeasured verdict)
carry_count absent                        after:  P10 FIRES (unknown treated as >0, never 0)
```

Gates on HEAD: gitleaks `rc=0` · layer-purity 0 violations · `validate-plugin.sh` PASSED · 8 checks green, 0 failing.

ekson73:
PDCA round applied in `122f877`. Two findings closed substantively, one declined with rationale — and the review was already **stale by ~3 minutes** when it landed (submitted `2026-07-29T23:47:31Z`; commits `5d5ba7f8` at `23:50:49Z` and `4576855f` at `23:51:41Z` followed it), so part of what it asked for was in flight as it wrote.

|---|---|---|---|
| 2 | `$LEDGER_DIR/<slug>.json` publish needs an exclusive lock; `.tmp` must be created inside `$LEDGER_DIR` | ✅ **CONFIRMED** | specified in §6.0 |
| 3 | resume/ledger validation must fail closed for per-run fields (`carry_count`, `dst_pinned_path`) | ✅ **CONFIRMED** | closed in §6.2 |
| 1 | shorten the v0.6.0 runtime changelog; move forensics to an external release note | ⚠️ **DECLINED** | rationale below |

**F2 — atomicity is NOT serialization, and that distinction was the gap.** The publish was already atomic (`.tmp`+`mv`, never `>`), but atomic ≠ serialized: two overlapping runs could each last-writer-wins the whole file and erase the other's `phase_states` / `tiers` / `impediments`. §6.0 now mandates read-modify-publish under an exclusive lock on `$LEDGER_DIR`. Your second half was the sharper one: the `.tmp` must be created **inside** `$LEDGER_DIR`, because `mv` is atomic only within one filesystem — staging in `/tmp` silently degrades to copy+unlink and the torn window returns one indirection out. Both are now pinned in the block.

**F3 — closed, and it surfaced a hole neither of us had named.** While validating the per-run fail-closed path I measured a distinct defect (trace 13): `RUN_ID` was assigned at its **old** site, which still reflected the narrower job it had *before* v0.6.0 promoted it to OWNER of the per-RUN fields. §6.0 mandates writing a phase entry BEFORE entering the phase — so a Phase-0 writer with `RUN_ID` unset persists `"run_id":""`, and a pre-assignment reader then compares `'' == ''`, **MATCHES**, and adopts another run's `drift_verdict` as its own with **P19 silent**. Set-vs-empty already failed closed; **empty-vs-empty** was the hole. Fixed with a `${RUN_ID:?…}` guard at process start.

Plus the rule your finding actually asked for: a ledger with **no `run_id` key at all** (legacy ≤v0.5.5, or written before `RUN_ID` was set) is an **UNATTRIBUTED** measurement ⇒ every per-RUN field in it is a MISSING MEASUREMENT. Same class as the legacy-impediment rule, same reason: absence of provenance is `unknown`, never *"mine"*. Two guards worth stating explicitly:
- it does **NOT** halt a resume — `carry_count` is per-CUTOVER and monotonic, so it is still inherited (§3.1), and its **absence** on an existing ledger fails **closed** (`unknown ⇒ >0`, keeping P10 armed) rather than defaulting to `0`, which would disarm P10 by making its condition someone else's and never false;
- ⛔ it is **never** back-filled with your own `$RUN_ID` — stamping it makes another run's verdict *read* as yours, which is the defect wearing the fix's clothes;
- `dst_pinned_path` absent or unreadable is likewise a MISSING MEASUREMENT ⇒ **P17 keeps blocking** until this run re-pins. Reading it as *"no destination ref"* is exactly the free-negative P17 exists to stop.

**F1 — declined, and the reason is the artifact's purpose.** The ask is to strip the forensic measurements, fixture details and reproduction history from the v0.6.0 changelog row into an external release note. Declined because in *this* artifact the forensics **are** the load-bearing content, not decoration:

- This is a **SKILL.md consumed by an amnesic agent**, not a library changelog for humans browsing releases. The row that reads *"measured both: after `git worktree remove`, per-worktree ledger DESTROYED, common-dir ledger SURVIVED"* is what stops the next agent from re-deriving `--absolute-git-dir` and re-breaking the amnesic contract a third time. Moving it to a release note the agent never loads converts a live guard into a dead file — the `[C08]` *"arquivo morto"* trap.
- The file's own history is the evidence: the same meta-failure (*the pattern was inherited, the fix was not*) recurred **three times** in this artifact. Each recurrence was prevented on the next pass only because the prior measurement was still in-context.
- There is a real cost and I am not pretending otherwise — the row is long and the file is ~140 KB. The right fix is **tiering** (one-line entries in the changelog, detail in per-topic sections), which is a structural pass across the whole document rather than a truncation of the newest row. Truncating only v0.6.0 would make the newest, least-internalized lesson the *thinnest* one. Tracked as follow-up rather than done half-way here.

Verified on `122f877`: diff is **additive** (+23/−0, no line removed) · fenced blocks balanced (20) · `validate-plugin.sh` PASSED · gitleaks clean. All checks SUCCESS except `code/snyk (ekson73@github)` = `Code test limit reached` — an account-quota **error**, not a finding (non-blocker per standing operator ruling 2026-07-12).

> ⚠️ Provenance note, stated because it affects how this commit should be read: the +18 lines closing F2/F3 were found **uncommitted in the worktree**, authored by an earlier session, not by me. Before committing I verified no peer was mid-write (no `index.lock`, no live git mutator, mtime ~2 h stale) and backed the file up first, per `script-safety` §1.1. I reviewed the content and stand behind it; I did not originate it. One instrument note while checking for a concurrent writer: `wc -c` (141888 **bytes**) and Python text-mode `len()` (139630 **characters**) disagreed by 2258 — that gap is the multibyte glyphs (⛔ · ·), not a second writer; hash matched the backup exactly.

@coderabbitai review

ekson73:
Round-2 finding applied in `a8d1982` — **valid, and it was a defect I introduced in `122f877`.**

The `${RUN_ID:?…}` fail-closed guard sat **ABOVE** the assignment, so it aborted unconditionally on every run and made the assignment **unreachable**. The guard written to close the empty-vs-empty hole would instead have halted the flow outright — the fix wearing the defect's clothes.

Fixed to generate-if-unset **first**, validate **second**, with `:-` so a caller-supplied value is preserved (a bare `=` would silently clobber it).

Verified by **executing** the pattern, not reading it — 3 cases + 1 negative control:

| case | result |
|---|---|
| unset | GENERATES (`29266-BC42E671`), rc=0 |
| `RUN_ID=caller-123` | PRESERVED verbatim, rc=0 |
| empty string | GENERATES — the empty-vs-empty hole this exists to close, rc=0 |
| **negative control**: old order (guard first) | `RUN_ID: RUN_ID must be set`, **rc=1** — confirming the assignment was genuinely dead code |

The order-is-load-bearing note is now in the file so the next reader does not re-invert it.

Verified: `validate-plugin.sh` PASSED · gitleaks clean · additive.

@coderabbitai review

ekson73:
## PDCA — 6 findings verified against this branch (`a4de767`)

Each finding was checked against the **branch code**, not the diff summary. **1 valid, 5 stale.**

### ✅ Applied — CodeRabbit ~1009 (changelog context pressure)

Valid, and measured: the changelog was **62,414B of the 142,272B file (44%)**. This file is
**instruction loaded into an agent's context**, so 44% of it was history displacing the
safety-critical P1–P19 trip-wires.

Extracted to `CHANGELOG.md` per `[C07b]` separate-spec-file (precedents: `harmonic-self-conduct-laws`
56KB→40KB, `auto-self-harness` 62KB→40KB). **SKILL.md 142,272B → 79,858B (−43%)**; 20 rows preserved
**verbatim** (append-only — nothing rewritten); P10/P17/P19 verified intact post-extraction (8/7/8 refs).

### ⏭️ Stale — the 5 below describe earlier commits

| # | Finding | Already in the branch |
|---|---|---|
| CodeRabbit ~523 | "Initialize `RUN_ID` before validating it" | L522-523 is already `RUN_ID="${RUN_ID:-$$-...}"` **then** `: "${RUN_ID:?...}"`, preceded by an explicit `⛔ ORDER IS LOAD-BEARING — generate-if-unset FIRST, validate SECOND` comment. The `:-` form also preserves a caller-supplied `RUN_ID`. |
| Copilot ~219 · qodo #1 | "`carry_count` treated as per-RUN but §3.1 uses it as cutover-level" | P19 (L219) already reads ``carry_count`` (per-**CUTOVER**, monotonic)`` and explicitly separates it from the per-RUN pair (`drift_verdict` · `dst_pinned_path`). The `run_id` schema comment repeats it: `⛔ carry_count is per-CUTOVER + monotonic ⇒ NEVER discarded (§3.1/P10)`. |
| Copilot ~744 · qodo #2 | "`run_id` owns fields 'below' but `drift_verdict` appears earlier ⇒ ambiguous" | The comment already reads `⛔ OWNER of the per-RUN fields — drift_verdict + dst_pinned_path — **REGARDLESS OF ORDER in this object**`. Positional wording (`below`/`above`/`earlier`) greps to **zero** in the file. |
| CodeRabbit ~788 | "Make missing per-run fields fail closed" | P19 already requires it: absence of `carry_count` `**read as 0** instead of **unknown ⇒ >0**` is a trip-wire, and `dst_pinned_path`-followed-as-address is another. P17: `An absent pinned sha is a **MISSING MEASUREMENT**, never "no destination ref"`. |
| CodeRabbit ~706 | "Serialize concurrent ledger updates + require same-filesystem publication" | Both already prescribed at L717-721: `⛔ the .tmp MUST be created INSIDE $LEDGER_DIR — mv is atomic only within one filesystem; from /tmp it degrades to copy+unlink` and `⛔ atomicity is NOT serialization: two overlapping runs can still last-writer-wins whole-file. Read-modify-publish under an **exclusive lock on $LEDGER_DIR**`. |

### Method note (own error, recorded)

My first pass on the ~706 finding used `grep ... | head -8` and the `exclusive lock` hit was #9 —
so I initially read the truncated output as *"zero hits ⇒ finding valid"* and nearly wrote a fix for
something already fixed. Re-ran without `head`: 3 hits, all present. Same class as the free-negative
this skill's own §5.1 notes warn about — **a truncated probe is not an absent result**.

Bots: please re-review `a4de767`.

ekson73:
@coderabbitai review

The `CHANGES_REQUESTED` verdict is pinned to `a8d1982`; HEAD is now **`a4de767`**. All 6 findings were addressed in the PDCA comment above — 1 applied (the ~1009 changelog context-pressure finding: `SKILL.md` 142,272B → 79,858B, **−43%**, history extracted verbatim to `CHANGELOG.md` per `[C07b]`), 5 verified stale against branch code with line-level evidence.

Note the `code/snyk` ERROR is `"Code test limit reached"` (plan quota, not a finding) — `security/snyk` is SUCCESS.

ekson73:
Round-3 applied in `f788a0f` — **valid, and it is the one class my v0.6.0 fix could not see**: I had just promoted `RUN_ID` to *owner* of the ledger's per-RUN fields and guarded it with `: "${RUN_ID:?…}"`, which proves the value **exists** and nothing about its **shape**.

The value is **caller-supplyable by design** (the `:-` preserves it so a resume can re-declare its identity) and is interpolated into **two sinks whose defences are not equal** — measured, not reasoned:

| sink | probe | result |
|---|---|---|
| **git ref** | `git update-ref "refs/rct/_dst/slug/$RUN_ID/refs/heads/main"` | `../../escape` · `x/../../y` · `a b` → **rc=128** (⇒ P16 clause 1 HALTs); `check-ref-format` **REJECTs** all three |
| **filesystem path** | §5.1's `DST_PINNED="$LEDGER_DIR/<slug>.dst-pinned.$RUN_ID.txt"`, verbatim | `x/../../../VICTIM/pwned` → **WROTE `VICTIM/pwned.txt` outside `$LEDGER_DIR`**; `../../../VICTIM/prod-secrets` landed a file beside a decoy secret |

So traversal **cannot** escape the ref namespace, and **can** write an arbitrary path. The ref sink being fail-closed is what made this invisible: I read one sink's safety as coverage for a sink that shares nothing with it.

⚠️ **The value that passes BOTH is the instructive one**: `-flag` created the ref *and* wrote the ledger file. It is not a traversal — it is an **argument-injection seed** for any later `git`/`rm`/`find` that takes the name positionally. Non-emptiness admits it; a format gate does not.

**Fix**: constrain to the shape the generator itself emits (`<pid>-<hex8>`) — `[A-Za-z0-9._-]{1,64}`, never a leading `-` or `.` — placed **after** the generate-if-unset (round-2's order-is-load-bearing lesson still holds: a guard above the assignment makes it unreachable) and **before** the first ledger read. Fails **closed**, because the alternative is a write whose destination the caller chose. **P19 extended** to name the format requirement.

**Verified on the block extracted verbatim from `SKILL.md` — not a retyped copy of it** (testing my reimplementation would prove only that I can write a working gate twice): **12/12** hostile REJECTED (traversal · space · control byte · `;` `$()` `|` · leading dash/dot · 65 chars) and **7/7** legitimate ACCEPTED, including **three live samples of the generator's own output** as the positive control that proves the gate is not simply refusing everything.

**The generalization, now in P19 and the changelog**: *"the variable is validated"* is not a property of the variable — it is a property of **each sink it reaches**. ⛔ **Enumerate the sinks, not the variable.** This is the fifth round on this PR where the fix was correct about the thing it was told about while inheriting the previous layer's unexamined assumption; the assumption this time was that one sink's rejection generalizes.

Boy-scout in the same commit: the `CHANGELOG.md` header said *"append no FIM"* while the table is **newest-first** — a changelog whose instructions contradict its own data teaches the next writer the wrong place. Corrected; row `0.6.1` prepended; `version` → `0.6.1`; row-count reference synced 20 → 21.

`validate-plugin.sh` PASSED · gitleaks clean.

@coderabbitai review

ekson73:
@coderabbitai — round 4. **Finding valid and fixed** in `261ab6e`. You caught an ordering defect that a diff cannot show, and the file itself was already confessing it.

## The finding, restated

> `RUN_ID` generation and shape validation are only shown in the §5.1 block, but §6.0 says Phase 0's **first** action is to read the ledger. That makes "validate before *any* ledger read" unreachable for a literal Phase-0 resume.

Correct. §5.1 is the **drift-detector** (pre-Phase-2); Phase 0 runs before it. So on a resume the gate I added in round 3 **never executes**, and the arbitrary-path write it closes is re-open on the exact path that matters.

⛔ **The file already admitted this and I shipped it anyway.** Line 512 read:

> `⛔ ASSIGN THIS AT PROCESS START — BEFORE the Phase-0 ledger read (§6.0), not here.`

…while the block stayed *there*. A comment naming its own violation is not a fix, and it survived a full review round because the code it annotates **looks correct in isolation**.

## Second consequence — the one that makes this more than a move

With `RUN_ID` unset at Phase 0, the writer persists `"run_id":""`, and a pre-assignment reader then compares `'' == ''` and **MATCHES** — adopting another run's `drift_verdict` as its own with **P19 SILENT**.

v0.6.0 closed **set-vs-empty**. **Empty-vs-empty** is a distinct hole that only *ordering* closes — no amount of format-validation reaches it.

## Fix

- New **§0.2 BOOTSTRAP** — the run's first executable step, ahead of every phase — holding the full `generate-if-unset → non-empty (:?) → format (case) → export` sequence, with the two-sinks rationale kept beside it.
- **§5.1 demoted to a pure consumer**, with an explicit ⛔ *do-not-re-derive*: a second `${RUN_ID:-…}` there would mint a **different** id than the one Phase 0 persisted, so the run would read its own earlier rows as **another run's** — self-inflicted amnesia wearing a safety check's clothes.

## Verification — extracted **verbatim from the file**, never from recall

```
hostile REJECT: 11/11   traversal · x/../../y · 'a b' · -flag · newline · glob* · ; · $(id) · 65-char · .hidden · a/b
legit  ACCEPT:  7/7     12345-abcdef01 · a · A.B_c-9 · resume.2026-07-30 · 0 · x_y.z-1 · 64-char boundary
caller-supplied valid   PRESERVED  (my-custom.id_42 → my-custom.id_42, not clobbered)
empty / unset           GENERATES  70697-AB6A68E1 · 70703-EB63830A · 70709-E9884E40  (3 distinct draws)
```

⚠️ **My own hostile set was wrong first.** I listed `RUN_ID=''` as a value that must be REJECTED and read its `rc=0` as a leak — but `''` is precisely the *unset* case the `:-` must **generate** for. The instrument was mis-specified, not the gate.

**Negative control** (the half that proves the gate is load-bearing) — a gate-stripped variant, with the strip *asserted* (`RUN_ID_BAD` absent) before trusting it:

```
nogate -> ESCAPE WROTE OUTSIDE ledger: .../deep/a/VICTIM/pwned.txt
gate   -> no write outside ledger
```

⚠️ That control was **also invalid on the first attempt**: I checked `$T/VICTIM` while `ledger/x/../../../VICTIM` resolves **above** `$T` entirely, so *both* variants reported "no write" and the defect looked absent — a **wrong-place** negative, from §5.0's own checklist, in the very PR that documents it. Re-run against the resolved path.

## The generalization (now in the changelog)

This is the **fourth ordering-shaped defect** in this skill's history: `carry_baseline` (write-once vs re-derive) · P18 (write-time vs read-time) · the ledger address (record vs address) · and now this. Each was a **correct check at the wrong moment**.

A guard's correctness has two independent components — *what it checks* and *when it runs* — and review reliably audits only the first.

⛔ **For every guard, name the sink it protects and prove the guard executes before that sink's first reachable use.** A diff cannot show ordering, so it must be asserted, and the assertion belongs beside the guard.

| Check | Result |
|---|---|
| `validate-plugin.sh` | ✅ PASSED · Errors: 0 · Warnings: 1 (pre-existing) |
| `gitleaks protect --staged` | ✅ no leaks found |
| Version | `0.6.1` → `0.6.2` (frontmatter + changelog row + newest-first range) |

Could you please re-review?

@coderabbitai review

ekson73:
## PDCA — 3 findings verificados contra o branch (`ed8cfff`)

Cada finding checado contra o **código do branch**, não o resumo do diff. **1 válido aplicado, 2 stale.**

### ✅ Aplicado — CodeRabbit ~RUN_ID entropy fallback (`3686366016`, Major, Data Integrity)

Válido. Reproduzido com as duas fontes forçadas a falhar: a pipeline chega em `head` com sucesso e produz `RUN_ID="$$-"` — passa no guard de não-vazio E no format gate do v0.6.1. Fix: sufixo capturado separadamente + abort quando vazio, antes do `:-`. Verificado nas duas direções: draw normal gera `45463-10B4053A`; falha forçada aborta rc=1 sem leak.

### ⏭️ Stale — CodeRabbit ~literal pipes CHANGELOG L20/L32 (`3686366008`, Minor)

Verificado contra o HEAD: as linhas citadas (0.5.3 em L20, 0.4.1 em L32) não contêm `||` literal. As tabelas renderizam corretamente. Nada a escapar.

### ✅ Aplicado (lado pointer) — CodeRabbit ~changelog ordering policy (`3686366026`, Minor)

O CHANGELOG.md header já havia sido corrigido no v0.6.1 ("newest-first"), mas o **pointer em SKILL.md** ("Append new versions at the END of that table") ainda contradizia a ordem real. Corrigido para "at the TOP (newest-first)".

---

Commits: `ed8cfff` (v0.6.3). Gitleaks clean. @coderabbitai review

ekson73:
@coderabbitai review

ekson73:
## PDCA — round 6, 3 findings verificados contra o branch (`090c2fd`)

**2 aplicados, 1 aplicado-via-peer.**

### ✅ Aplicado — entropy only when RUN_ID unset (`3693716660`, Major, Stability)

Válido — e era um **defeito que minha própria fix v0.6.3 introduziu**: `_RUN_ID_SUFFIX=$(…)` rodava incondicionalmente antes do `:-`, então com entropia falha um caller com `RUN_ID` válido abortava. Reproduzido (uuidgen+od forçados off → FATAL mesmo com caller). Fix: entropia só quando `RUN_ID` unset; check vira `[ -n "$RUN_ID$_RUN_ID_SUFFIX" ]`. Verificado 4/4 combinações (no-caller+OK gera · no-caller+FAIL aborta · caller+FAIL **preservado** · caller+OK preservado).

### ✅ Aplicado — §0.2 explicitly never-skippable (review body, Major, Security)

Válido. A safety-floor list nomeava T4 · §9.0 · §4 · §5.0–5.2 · §3.1 · Phase-4, mas **não §0.2** — um skip §0 alcançaria o Phase-0 ledger read com `RUN_ID` não validado, reabrindo exatamente as falhas que §0.2 existe para prevenir. Adicionado como primeiro item da lista never-skippable.

### ✅ Aplicado (peer) — filename `CHANGELOG.md` → `changelog.md` (`3693716658`, Minor)

Regra real (`AGENTS.md:34` lowercase-hyphenated). Inicialmente avaliei como defer (blast estimado alto); uma sessão paralela aplicou corretamente em `9d8aa6e` — o blast real era 2 refs + rename de mesmo conteúdo. Minha avaliação de custo estava errada; registrado.

---

Commits: `ed8cfff` (v0.6.3) · `9d8aa6e` + `e3026f0` (v0.6.4) · `090c2fd` (wording). Gitleaks clean. @coderabbitai review

PR #297, opened 2026-07-30
feat(research-dossier): the declared form stops lying, and the page fits a phone (v0.2.0)

ekson73:
## What this is

`research-dossier` v0.1.0 → v0.2.0. Two gaps found by reading v0.1.0 against what callers assume it does.

The originating request asked to *forge* a multi-format dashboard system. Recon found it already shipped — `1a35fdf` (#288), the same day. So this is a version bump, not a sibling skill: `scope-discipline` Q2 (≥50% covered ⇒ EXTEND) and no second SSOT.

## 1. `FORM_NOT_RENDERED` — new WARN, `--strict` ⇒ FAIL

The schema accepts **13** `chart.form` values. The renderer draws **one** geometry: horizontal proportional bars (`<rect>` at `render.mjs:797` + baseline `<line>` at `:811`). Measured: the renderer branches on `form` *only inside the gate* (`:245`), never in rendering.

So `form:"line"` was accepted, ignored, and silently substituted — the IR says line, the artifact shows bars, nothing announced it. That is theater of exactly the kind this skill exists to refuse, **inside the skill built to refuse it** (anti-theater Layer-5).

WARN rather than FAIL is proportionate: the evidence stays sourced and truncation-checked; only the label overpromises.

### Two alternatives rejected on measurement, not taste

| Option | Why not |
|---|---|
| Shrink the enum | **BREAKING.** `form` is `enum`-constrained with `additionalProperties:false` ⇒ any existing IR carrying `form:"line"` stops validating (MAJOR, not MINOR). Also deletes vocabulary the bundled `dataviz` reference actively teaches — the schema itself says *"stat-tile / kpi-row / table are first-class answers there"*. |
| Build 13 geometries | **YAGNI.** 6 of 6 fixture charts declare only `"bar"` ⇒ 12 of 13 forms have **zero** test coverage. |

### ⚠️ This adds no exemption to anything

The `MAGNITUDE_FORMS` allow-list is **not** back. Truncation stays form-agnostic, exactly as `render.mjs:88-95` records its deliberate removal — *"so the next reader does not reintroduce the allow-list as an apparent improvement"*. I nearly made that error: my first analysis claimed this change would "reopen a gate hole". Wrong — the bypass was already closed, and `FAIL` msg `:224` says so out loud (`declared form "X" does not exempt it`). The real risk is narrower and named in the code comment: the `keys off what is DRAWN` predicate expires the moment forms gain real geometry.

## 2. Width breakpoints (760px / 420px)

Before this, the only `@media` queries were `prefers-color-scheme`, `prefers-reduced-motion`, `print`. A viewport meta plus a `max-width` made the page **shrink**, which is not the same as readable. A decision artifact is read where the decision happens, often not at a desk.

Header stacks; wide data tables scroll instead of the page (the table *is* the accessible rendering — it must never be clipped); claim ids wrap. Declared **before** `@media print` so print keeps the last word.

## 3. Stated, not implied

New "Not built" table in SKILL.md — each **probed**, not guessed: drill-down (absent) · filter controls (declared as `dataviz` parameter #8 in `audience-map.md`, inert) · burn-down (absent, no consumer — deferred rather than built on spec) · stat-tile demotion (self-declared "Not yet implemented") · yaml (verified: `--formats yaml` ⇒ exit 0, `UNKNOWN_FORMAT` surfaced, **zero files written**).

## Verification

- **93 passed, 0 failed** (was 89). Every new assertion mutation-verified in **both** directions.
- **One assertion was caught being a test that cannot fail.** It addressed the fixture through a `$TMP`-relative path that silently `ENOENT`s ⇒ node printed no code ⇒ the negative `case` matched ⇒ `ok()` fired unconditionally. It would have passed with the guard deleted **and** inverted. Fixed to address `$EX` directly, plus an empty-output branch so an assertion can never again pass by observing nothing.
- Guard neutered ⇒ suite 92/1 ✗; restored ⇒ 93/93 ✓.
- Idempotent render verified (`diff` of two runs identical) · `tests/validate-plugin.sh` **PASSED** · zero npm deps · zero remote subresources · JS-off render unaffected · 6 documented fixtures unchanged (0,1,1,1,1,1) · `v0.2.0` stamped into the artifact.

## Not done (deliberate)

Real pdf/pptx/xlsx documents. The owners are installed (`~/.claude/plugins/cache/anthropic-agent-skills/document-skills`, `marketplaces/open-design/skills/pptx`) and `od` is on PATH, so the last-mile is reachable — but it is a separate concern with its own failure modes, and it does not belong in a gate-integrity PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ reference teaches; building 13 geometries is YAGNI while 6 of 6 fixture charts d
 only `bar`, leaving 12 forms with zero test coverage.
 
 ⚠️ This adds **no exemption** to anything. The `MAGNITUDE_FORMS` allow-list is *not*
-back: truncation stays form-agnostic, exactly as `render.mjs` records above
+back: truncation stays form-agnostic, exactly as `bin/research-dossier-render.mjs` records above
 `gateProvenance` ("so the next reader does not reintroduce the allow-list as an
 apparent improvement"). The new code only reports the declared-vs-drawn gap.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed — `research-dossier` v0.2.0: the declared form stops lying, and the page fits a phone
+
+Two gaps found by reading v0.1.0 against what callers assume it does.
+
+**`FORM_NOT_RENDERED` (new WARN, `--strict` ⇒ FAIL).** The schema accepts 13
+`chart.form` values; the renderer draws exactly one geometry — horizontal proportional
+bars (`<rect>` + baseline `<line>`). So `form:"line"` was accepted, ignored, and
+silently substituted: theater of precisely the kind this skill refuses, inside the
+skill built to refuse it. The gap is now named and deterministic. WARN rather than
+FAIL is proportionate — the evidence stays sourced and truncation-checked, only the
+label overpromises.
+
+Two alternatives were rejected **on measurement, not taste**: shrinking the enum is a
+breaking change (`enum` + `additionalProperties:false` ⇒ an existing IR carrying
+`form:"line"` would stop validating) and would delete vocabulary the bundled `dataviz`
+reference teaches; building 13 geometries is YAGNI while 6 of 6 fixture charts declare
+only `bar`, leaving 12 forms with zero test coverage.
+
+⚠️ This adds **no exemption** to anything. The `MAGNITUDE_FORMS` allow-list is *not*
+back: truncation stays form-agnostic, exactly as `render.mjs` records above
+`gateProvenance` ("so the next reader does not reintroduce the allow-list as an
+apparent improvement"). The new code only reports the declared-vs-drawn gap.
+
+**Width breakpoints (760px / 420px).** Before this the only `@media` queries were
+`prefers-color-scheme`, `prefers-reduced-motion` and `print` — a viewport meta plus a
+`max-width` made the page *shrink*, which is not the same as being readable. A decision
+artifact is read where the decision happens, often not at a desk. The header stacks,
+wide data tables scroll instead of the page, claim ids wrap. Declared before
+`@media print` so print keeps the last word.
+
+**Stated, not implied** — a new "Not built" table in SKILL.md records what is absent
+and why, each probed rather than guessed: drill-down (absent), filter controls
+(declared in `audience-map.md` as `dataviz` parameter #8, inert), burn-down (absent, no
+consumer — deferred rather than built on spec), stat-tile demotion (self-declared "Not
+yet implemented"), yaml (unknown format ⇒ logged `UNKNOWN_FORMAT`, verified to write
+zero files).
+
+- **93 tests pass** (was 89). Every new assertion mutation-verified in **both**
+  directions. One of them was caught being a *test that cannot fail*: it addressed the
+  fixture through a `$TMP`-relative path that silently `ENOENT`s, so node printed no
+  code, the negative `case` matched, and it would have passed with the guard deleted
+  *and* inverted. Fixed to address `$EX` directly, plus an empty-output branch so an
+  assertion can never again pass by observing nothing.
+- Idempotent render verified; `tests/validate-plugin.sh` PASSED; zero npm deps; zero
+  remote subresources; JS-off render unaffected.
+
 ### Added — `research-dossier`: finished research → decision-ready visual dossier, behind two deterministic gates (#287)
 
 The missing spine between researchers and the ~110 `design-templates/` + ~150 `design-systems/`
--- a/bin/research-dossier-render.mjs
+++ b/bin/research-dossier-render.mjs
@@ -973,7 +973,11 @@ function main() {
   --formats <list>     html,md,json          [default: html,md,json]
   --out <dir>          output directory      [default: out]
   --gates-only         run both gates, render nothing
-  --strict             a missing dataviz validator FAILS instead of warning
+  --strict             escalate advisory findings to build failures:
+                       · a missing dataviz validator (gate 2) FAILS instead of warning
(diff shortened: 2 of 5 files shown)


ekson73:
@qodo-code-review Both findings ran against the input rather than judged by reading. One landed, one is declined with evidence.

| # | Finding | Disposition | Commit |
|---|---|---|---|
| 2 | `--strict` help outdated | **FIXED** | `956f063` |
| 1 | literal CSS values instead of `var(--token)` | **Declined — no such tokens exist** | — |

### #2 — correct, and it was mine

Verified by running it:

```
$ node bin/research-dossier-render.mjs --help | grep strict
  --strict             a missing dataviz validator FAILS instead of warning
```

I widened `--strict` to also escalate `FORM_NOT_RENDERED` and never updated the help. A CLI whose help understates when it fails is a CI trap — exactly right.

Fixed in **both** places the drift existed. Your pointer named `--help`; checking the rest of the surface found `SKILL.md:106` carrying the same stale sentence in the flags table. The command surface (`commands/research-dossier.md`) and the degradation prose (`SKILL.md:204`) were checked and are still accurate — they scope their claim to the validator, which remains true.

### #1 — declined, with the probe

The tokens the policy asks me to use **do not exist in this template**. Every token defined in `:root` is a *colour* token:

```
--page --surface-1 --text-primary --text-secondary --muted
--border --grid --baseline --good --success-text
```

There is no `--space-*`, no `--radius-*`, no `--font-size-*`. Probed directly:

```
$ grep -oE '\-\-(space|spacing|radius|font|size)[a-z-]*:' skills/research-dossier/templates/dossier.html
(no output)
```

And the pre-existing rules I did not touch use literal px on the same properties:

```css
.wrap  { max-width: 1080px; padding: 32px 20px 72px; }
.card  { border-radius: 12px; padding: 20px 22px; }
h1     { font-size: 26px; }
```

So converting only my breakpoints to `var(--token)` would mean **inventing** a token vocabulary that exists nowhere else in the file, leaving two competing conventions in one stylesheet — measurably worse maintainability than the single consistent one now present. My breakpoints deliberately mirror the surrounding style.

Introducing a spacing/radius/type scale here is a real improvement, but it is a whole-template refactor touching every rule above line 150 — not a gate-integrity PR. If you want it tracked I will open it separately rather than half-apply it here.

Re-review requested.

ekson73:
@coderabbitai @copilot-pull-request-reviewer Every finding was probed against the code, not judged by reading. Three landed, one is declined with a measurement.

| Bot | Finding | Disposition | Commit |
|---|---|---|---|
| Copilot | width breakpoints leak into `@media print` | **FIXED** — and it falsified a claim in my own comment | `ee34154` |
| Copilot | `render.mjs` path matches no file | **FIXED** | `ee34154` |
| CodeRabbit | `--strict` help incomplete | **FIXED** (you auto-detected it: `✅ Addressed in 956f063`) | `956f063` |
| CodeRabbit | sign documents with agent ID + timestamp | **Declined — zero precedent, measured** | — |

### Copilot #1 — the best finding on this PR

It caught me asserting something I had not verified. My comment said the breakpoints were safe because they sit before `@media print`, so *"print keeps the last word"*. Probed:

```
print block (real line 185) → overflow-x rules: 0
                            → max-width rules:  1   (control)
```

Print resets `.wrap` but **never touches `overflow-x`**. An unqualified `(max-width: 760px)` also matches print, whose page box is narrow — so `details.tableview { overflow-x: auto }` stayed live on paper, where nothing can scroll, clipping exactly the wide tables the print block exists to render in full. The accessible rendering of every chart *is* that table, so this was the worst place to be wrong.

Source order only protects properties print actually overrides. Ordering protects those; `screen` protects the rest. Both queries are now `@media screen and (max-width: …)`, verified **in the rendered artifact** rather than the template: 2 screen-scoped, 0 unscoped leftovers.

My first probe of this was itself wrong — a `sed` range anchored on `/@media print/` matched a *comment* at line 28 and swallowed my own breakpoints, reporting `overflow-x: 2` and making the bug look absent. Re-probed with real line bounds. The instrument lied before the code did.

### CodeRabbit — signature declined, with the count

The convention is real (`CLAUDE.md`: *"Sign documents with agent ID and timestamp"*), but it has **no precedent in either file this PR touches**:

```
agent-id signatures in CHANGELOG.md:              0     (control: 'research-dossier' → 6 ✓)
signatures in skills/research-dossier/SKILL.md:   0
sampled other SKILL.md (red-team, preflight,
  anima, postflight):                             0 each
```

114 CHANGELOG headings, zero signed. Signing only my entry would make this PR the sole outlier in a file that has never carried them, and would imply a convention the repo does not follow for changelogs or skill docs.

Worth flagging honestly: my *first* probe here appeared to find a precedent — it matched the prose *"the emitter signature"* in the v0.1.0 entry, not a signature. Re-probed with an agent-id pattern: 0.

The artifacts that *are* signed are the ones the guideline targets — rendered dossiers carry `research-dossier v0.2.0` (`RENDERER_ID`, asserted by a test and bumped in this PR). If changelog/skill-doc signing is wanted, that is a repo-wide convention change across 114 entries, not something to start unilaterally inside a gate-integrity PR.

**State:** 93/93 pass, `validate-plugin.sh` PASSED, idempotent, zero npm deps, 6 documented fixtures unchanged. Re-review requested.

ekson73:
@coderabbitai review

Your `CHANGES_REQUESTED` is pinned to `5dfcb75`; HEAD is now `ee34154` (two commits later).

- **`--strict` help** — you already auto-detected this one as `✅ Addressed in commit 956f063`.
- **Document signing** — declined with a measurement, not an opinion: `0` agent-id signatures across 114 CHANGELOG headings (control: `research-dossier` → 6 hits, so the probe reached the file), and `0` in `skills/research-dossier/SKILL.md` plus the four other SKILL.md I sampled. Signing only this entry would make it the sole outlier. Rationale in the comment above.

Also landed since your review: both width breakpoints scoped to `screen` (Copilot caught them leaking into `@media print`, where `overflow-x: auto` would clip the very tables the print block exists to render in full), and a bare `render.mjs` path corrected.

93/93 pass · `validate-plugin.sh` PASSED · idempotent · 6 documented fixtures unchanged.

PR #321, opened 2026-08-14
feat(skills): refine-braindump-to-prompt (Lapidary) — braindump → one executable prompt

ekson73:
**[PR-as-Prompt]**

## Context

The operator has a recurring ritual: dump raw intent — mixed goals, cartesian resource lists, session wrappers — and expect it to become executable work. That ritual breaks at a measured gap.

Three independent Explore sweeps (user-scope corpus, this repo's skill cluster, the `/enhance` cluster) converged on the same verdict: **no tool produces one executable prompt from one braindump.**

| Existing terminus | Produces |
|---|---|
| `directive-braindump-triage` | a classification ledger |
| `agentic-tool-forge` | a new tool |
| `braindump-distill` (vault) | vault artifacts |
| `/enhance` | a single pass, no loop |
| `enhance-pipeline` | a PR |

Every one ends in a JSON envelope, a ledger, or a PR. **Never a rendered prompt.**

## Objectives

Close that gap with **one** generic, parameterized skill — and make `gauntlet-loop` a reusable profile rather than a second skill.

- `skills/refine-braindump-to-prompt/` — RECOVER → DRAFT → REFINE → RED-TEAM → RENDER
- `profiles/default.md` — neutral shape
- `profiles/gauntlet-loop.md` — the Gauntlet Loop architecture, with attribution
- `commands/refine-braindump-to-prompt.md` — thin command surface

### The doctrinal contribution: the Verifiability Gate

Three sources disagree on when an agentic loop may stop:

| Source | Stopping rule |
|---|---|
| Shumer / Duola | *"loop until utterly perfect — you are the brake"* → **none** |
| Osmani (*Loop Engineering*) | 5th component, **mandatory** (cost · iterations · clock · stagnation) |
| This repo (`convergence-engine`) | `n* ≤ 3-4`; floor = human parity ~90%, **not** 100% |

The synthesis encoded in the skill:

```text
long_loop_licensed := bar_is_machine_verifiable
                    AND critic_is_independent_of_builder
                    AND evidence_is_directly_inspectable
```

A long loop is licensed **only** when the gain signal is real. Otherwise the economic stop applies. Unbounded is refused in every case.

## Acceptance

| Check | Result |
|---|---|
| `tests/validate-plugin.sh` | ✅ 0 errors, 1 warning — 85 SKILL.md files pass frontmatter |
| `gitleaks protect --staged` | ✅ no leaks (~33.5 KB + 10.2 KB) |
[...]

PR #322, opened 2026-08-14
fix(governance): rescue #321 unpushed root-cause — YAML-parse frontmatter validation

ekson73:
The merge of #321 (755ca20) captured the pushed head; four files of
post-push work remained uncommitted in the feature worktree. Rescued,
not dropped (NO-LOOSE-ENDINGS):

- tests/validate-plugin.sh: root-cause fix — frontmatter must PARSE as
  YAML, not merely grep-match. A grep answers 'does the text contain?',
  never 'does the parser accept?'. Capability-detected: PyYAML absent
  => WARN, never a red build for a missing dep.
- skills/refine-braindump-to-prompt/SKILL.md: its own frontmatter was
  the instance of the bug class (plain scalar with ': ') — now a
  folded block; n* notation normalized.
- CHANGELOG/EVAL-REPORT: precision wording + eval signature line.

Local run: validate-plugin.sh EXIT=0, new check PASSED with PyYAML
present (not skipped). Unsigned per [C19] §6.1 (signing agent down).

Co-Authored-By: Codex Opus <noreply@anthropic.com>

PR #323, opened 2026-08-14
fix(skills): P0 — refine-braindump-to-prompt frontmatter does not parse (+ close the validator hole that let it ship)

ekson73:
## P0 — the skill merged in #321 does not load

`skills/refine-braindump-to-prompt/SKILL.md` on `main` **fails YAML parsing**:

```
mapping values are not allowed here
```

The frontmatter `description` is a plain scalar containing `: ` (in `Five phases:` and
`--output-target: stdout`). Measured on `origin/main` with a positive control — `skills/quiesce/SKILL.md`
parses from the same tree, so the instrument is sound and the defect is real.

**Timing**: CodeRabbit flagged it 🔴 Critical on #321 at `2026-08-14T18:42Z`. The fix was committed to
the branch at `c8589fd`, ~minutes after #321 merged at `20:54:37Z` — so the merge captured `ffe1e7f0`,
the commit *before* the fix. Nobody's error; a race.

## What this PR does

| | Change |
|---|---|
| **P0** | `description: >-` folded scalar. The `: ` survives **inside** the value — syntax corrected, content not sanitized away. |
| **Root cause** | `tests/validate-plugin.sh` printed `✓ PASSED` on the broken file. Every frontmatter check greps for a **line** (`^name: x$`), which answers *"does the text contain?"* — never *"does the parser accept?"*. Fixing only the file leaves the next skill free to ship the same class. |
| **New gate** | YAML-parse check in the validator: capability-detected (PyYAML absent → WARN, never a red build for a missing dep — the pattern the script already uses for `python3`/json), skips `README.md` per the existing artifact/doc exclusion. |

## Why the gate is safe, measured before adding it

- **144** artifacts parse today; the only **3** without frontmatter are `README.md` / `COWORK-AUTONOMY-POLICY.md` — docs, already excluded. **No pre-existing breakage is turned red.**
- **Negative control**: reintroducing the exact shipped defect turns the build **red** with the exact message; restoring returns it **green**. A gate proven to bite, not asserted to bite.

## Also (CodeRabbit minors from #321)

- `n* ≤ 3-4` read as **subtraction** in math syntax → `n* (typically 3 or 4)`, plus the closed form it derives from (`agentic-first §4.7.6`) now stated, so the bound is a computed predicate rather than a magic literal.
- CHANGELOG population wording restored to the eval's measured *"stop/refuse/escalate points"*.
- Eval report signed (agent ID + ISO 8601) per the repo MUST.

## Verification

```
validate-plugin.sh          rc=0  ✓ PASSED
frontmatter parses          ✔ (PyYAML, keys: name/version/description/allowed-tools)
negative control            ✔ red on defect, green on restore
machine envelope            ✔ valid JSON after comment-strip
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,17 @@ reach, because it describes the rule without being it. Both the eval and its del
 produced a comparable prompt. The skill's measured delta is the mechanically-derived material and the
 independent red-team, not the qualitative judgment.
 
+v0.3.0 has since been re-evaluated by four blind agents against a frozen artifact — the independent
+re-eval the report itself asked for. Result: **3 of 5 claims held, 2 disproved, and 22 decisions the
+document does not make for a cold agent.** The floor repair is confirmed on both methods (a reader
+enumerating termination passages, and a run that exited REFINE at 3 rounds where an unconditional
+floor needs ≥6). Disproved: the referent halt site is assigned to two different phases by two
+passages of the same section; and 13 of 29 stop points carry no terminal marker, against the
+document's own assertion that a halt without one is an unfinished rule — 11 of those 13 outside the
+section that makes the assertion. Three inputs of three classes produced three halts and zero
+rendered prompts. The verdict stands at FAIL, but the failure is now located: **specification, not
+mechanism.**
+
 
 ### Added — squash-aware release-coherence gate
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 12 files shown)


ekson73:
Closing as a **duplicate of #322** — measured, not assumed.

`diff` of all four touched files between this branch and `fix/rescue-321-leftover`: **0 lines** on
each (`SKILL.md`, `tests/validate-plugin.sh`, `CHANGELOG.md`, `EVAL-REPORT-2026-08-14.md`). Same P0
fix, same root-cause validator gate, byte-identical.

The difference is base, not content: #322 branched from `main` **after** #321 squash-merged, so it is
`MERGEABLE`; this branch still carries the pre-squash history and is `CONFLICTING`. #322 is the
better-positioned carrier of the same work.

`compose-not-fork`: two PRs fixing one P0 is the duplication the contract exists to prevent. #322 lands it.

PR #324, opened 2026-08-14
fix(skills): retract a false n* claim — cite convergence-engine as SSOT

ekson73:
## Retracting a false claim I shipped in #322

While fixing a *different* CodeRabbit finding on #321 (`n* <= 3-4` reads as **subtraction** in math
syntax), I resolved the notation by asserting the closed form *"grows only logarithmically in `V/C`,
so it evaluates to **3 or 4** across realistic inputs."*

**That assertion is false, and I never computed it before writing it.** CodeRabbit contested it on
#322 (🟠 Major). I verified independently rather than accepting it on authority:

`n* = 1 + ⌈ln((1−ρ)·g₀·V/C) / ln(1/ρ)⌉`

| ρ | g₀ | V/C | n* | |
|---|---|---|---|---|
| **0.9** | 1 | 100 | **23** | CodeRabbit's counter-example — reproduces exactly |
| 0.5 | 1 | 100 | 7 | |
| 0.7 | 1 | 50 | 9 | |
| 0.3 | 1 | 100 | 5 | |
| 0.5 | 1 | 8.2 | 4 | the only case landing in {3,4} — and it was constructed to |

Five sampled parameter sets; none in `{3,4}` except the one built to be. `ρ=0.9` — a slow-converging
loop retaining 90% of the gap — is entirely realistic and yields **23**. The claim does not survive
its own arithmetic.

## The repair is a citation, not a better adjective

Per CodeRabbit's own prescription and the SSOT discipline **ratified today** for `--output-target`
(`coexistence-pr11xpr12-20260814`): **`skills/convergence-engine/SKILL.md` is the SSOT** for the
bound, its parameter ranges, its marginal-value indexing, and its round-count convention — *is `n*`
the last affordable round, or the first rejected one?* (the `A=4` boundary yields `n*=3` while round 3
is still affordable). This skill **cites** it and never restates it.

**8 occurrences** replaced · **0** residual literal cap claims · `validate-plugin.sh` **PASSED**.

## Deliberately out of scope

Defining the parameter bounds and the counting convention **inside** `convergence-engine` is the
complete fix. It touches another skill's SSOT and deserves its own cycle with a red-team — not a
doc-only retraction PR. Tracked as follow-up.

## The pattern, named

This recurred **three times** in one session: *fixing a symptom by adding a claim I had not measured.*
The notation ambiguity was real; the number I reached for to resolve it was not measured. **Citing
beats asserting** — and the reviewer who caught the P0 was right again about the fix for it.

### DIFF
--- a/skills/convergence-engine/SKILL.md
+++ b/skills/convergence-engine/SKILL.md
@@ -8,7 +8,7 @@ description: |
   three regimes by verifiability — REFINE (self-improve loop), SELECT (best-of-N /
   debate→converge), DEFER (HITL) — under a non-negotiable master condition
   (verifier > generator, verifier independent) and a closed-form economic stop
-  (robustly ≤3–4 rounds). Composes existing primitives; builds no new engine.
+  (bounded by an economic stop whose round count is dominated by the retained-gap fraction ρ — see Economic stop). Composes existing primitives; builds no new engine.
   Use when a result must be driven to high quality through bounded iterative review
   rather than a single pass, when multiple proposals must be reconciled, or when an
   autonomy_score must be lifted across the HIGH gate before escalating to a human.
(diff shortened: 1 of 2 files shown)


ekson73:
## PDCA — 3 findings, all verified before fixing

@chatgpt-codex-connector **F1 (P2) was the real one, and you were right.** I measured your claim rather than accepting it:

```
convergence-engine/SKILL.md:  'parameter bound' 0 · 'marginal' 0 · 'round-count convention' 0
                              'robustly <=3-4 rounds' asserted TWICE (:11 and :91)
```

**I had not removed the falsehood — I moved it**, into the document I then pointed at as authority. Delegating to an empty SSOT is worse than the literal, because the literal was at least checkable.

Populated the SSOT with what the delegation promised, **derived, not asserted**:

| | Now defined in `convergence-engine` |
|---|---|
| **Round-count convention** | Round *k* buys `V·g₀·(1−ρ)·ρ^(k-1)`; loop while > `C`. Solving gives **`n* = last_affordable + 1`** — verified over **16 parameter sets (15/16**; the 16th degenerate). ⇒ **`n*` is the FIRST REJECTED round**; the last worth running is `n*−1`. **Anyone treating `n*` as a budget runs one round too many** — a real bug the missing convention was hiding |
| **Parameter bounds** | Dominated by **ρ**, not `V/C`. The `ln(V/C)` numerator *does* grow logarithmically — that half was true, which is why the claim survived — but the **`ln(1/ρ)` denominator collapses as ρ→1**. Measured table shipped: ρ=0.3 → 4–7 rounds · **ρ=0.9 → 8–45** |
| **Marginal-value indexing** | Round *k* affordable iff its marginal value exceeds `C` — the gap that round **closes**, not the gap remaining. `k` starts at 1 |
| **Parameter-free escape** | If ρ can't be estimated, **don't quote a round count** — use the harness cap + stagnation test, which need no estimate |

@qodo-code-review both yours confirmed and fixed:

- **F2 (Medium)** — correct: my `replace()` substituted *into* existing code spans → nested backticks, 7 sites. Fixed, including the two inside code fences where backticks must not appear at all.
- **F3 (Low)** — correct: the cell claimed *"cites and never restates"* while restating the closed form in the same sentence. Restatement removed; the claim is now true.

```
frontmatters parse ✔ (both)   ·   validate-plugin.sh PASSED   ·   head 90525b5
F1 SSOT content: convention 1 · indexing 1 · bounds 1     F2 nested: 0     F3 restatements: 0
```

Re-review please.

PR #325, opened 2026-08-14
docs(skills): render the Gauntlet target — maos skill corpus vs anthropics/skills

ekson73:
## [PR-as-Documentation-Prompt]

**Context.** The operator answered the one open decision from the `/quiesce` closure — *"executar o
prompt Gauntlet contra um projeto real precisa de um alvo nomeado"* — with **B) maos itself
(recursive dogfood)**. But `maos` is a *place*, not *work*, and the rendered prompt's own
`ESCALATION` refuses a TARGET that is "absent or ambiguous". So the TARGET was derived from the
repo's measured state rather than bounced back as a second question.

**Objective.** Persist the rendered Gauntlet prompt as an armed-ready artifact. **This PR does not
run the loop.**

### What was measured (not assumed)

| Measured | Value |
|---|---|
| Agentic artifacts | 147 — 82 skills / 28 agents / 37 commands |
| `validate-plugin.sh` gates | structure only (frontmatter, JSON, exec-bit, token budget) |
| Nothing gates | quality vs an external reference |
| Bar half 1 | `anthropics/skills` — 169,418 stars, 17 refs, pushed 2026-08-13 |
| Bar half 2 | `agentskills.io/specification` — HTTP 200, 414,136 bytes |

Both bar probes ran with a **negative control** (a fabricated repo and URL returned 404 under the
same command), so the positives are discriminating reads, not an instrument that says yes to
everything.

### The trap that probing caught

`anthropics/skills` ships `spec/agent-skills-spec.md` at **87 bytes** — it is a *redirect stub*
(`The spec is now located at <https://agentskills.io/specification>`), not a spec. Naming that path
as the quality bar would have handed the blind critic a **pointer instead of a specification**,
failing the profile's own gate #2 (*the agent cannot observe the true result*). The byte-size was
the only tell; the path name reads perfectly.

### Honest gate declaration

The Verifiability Gate is recorded as **hybrid**, not pure-deterministic:

- **D** frontmatter parses (already gated by `validate-plugin.sh`)
- **T** description length, body word-count, progressive-disclosure presence — countable
- **J** blind A/B on clarity — judged

Marking this ✅-deterministic would be the exact theater the skill exists to refuse, so it is
declared instead.

### Why it ships un-armed

`profiles/gauntlet-loop.md` marks `cost_ceiling` and `wall_clock` as `operator-set; REQUIRED before
any long run` ("unbounded time is not a budget"), and the rendered `STOP` restates it: *"Unbounded
running is REFUSED — if either cap is unset, stop and ask."* The operator was asked and the caps
remain unset, so **the loop is refused by its own contract**, not by hesitation. Two slots stay open
for the operator; the recommended first run is a 3-skill pilot so a harness defect is discovered
across 3 pieces rather than multiplied by 82.

### Scope boundary

`agents/` (28) and `commands/` (37) are explicitly **out of scope** — Anthropic's corpus contains no
comparable artifacts, so extending this bar to them would name a reference that does not exist: the
same failure the `spec/` stub nearly caused.

### Acceptance

- [x] `tests/validate-plugin.sh` → PASSED (rc=0 captured **unpiped**)
- [x] `gitleaks dir .` → no leaks (7.29 MB scanned)
- [x] Worktree per `[C04]`; docs-only; single file added
- [x] Every bar claim carries a probe result + a negative control
- [x] Artifact ships un-armed; no budget consumed

### Cross-links

- `skills/refine-braindump-to-prompt/profiles/gauntlet-loop.md` (the profile rendered)
- `skills/refine-braindump-to-prompt/examples/DOGFOOD-gauntlet.md` (the meta-prompt with the `TARGET` slot)
- Follows #321 (skill), #322 (validator root-cause), #324 (n* SSOT retraction)

### DIFF
--- a/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
+++ b/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
@@ -4,7 +4,11 @@
 > **maos itself** as the target (recursive dogfood). This file is the **artifact**; it is not armed.
 > See §Budget — the loop is refused until a ceiling exists.
 >
-> Signed: **Claude-Doc-ef60-325** · rendered `2026-08-14T22:03Z` · revised `2026-08-14T22:58Z`
+> Signed: **Claude-Doc-ef60-325** · first rendered `2026-08-14T22:03Z`
+> · last revised: **`git log -1 --format=%cI -- <this file>`** — read it, do not copy it here.
+> A hand-written `revised` stamp lived here and went stale three times in one session,
+> silently, because it duplicates a fact git already tracks authoritatively. The signature
+> the review asked for is the agent ID; the timestamp is git's to answer.
 
 ## Why this target, from measurement
 
@@ -188,6 +192,14 @@ cannot tell. It may not lower the bar.
 Run bin/convergence-guard for what it does catch (correlated brand/axis peers). Note its
 limit: it compares caller-supplied labels and cannot observe context freshness.
 
+  NOT RUNNER-ENFORCED — stated, not hidden. A review asked for critic isolation to be a
+  runner-enforced contract. It is not, and no primitive in this repo can make it one:
+  convergence-guard reads labels the CALLER supplies, so a caller that mislabels a
+  context-contaminated critic passes the guard. Isolation here is a PROMPT-level contract
+  the invoking agent can violate without detection. Closing it needs a runner that spawns
+  the critic itself and controls what enters its context — which does not exist yet.
+  Recorded as a known hole rather than described in language that implies a gate.
+
 EVIDENCE  (directly inspectable — no inference from the build story)
   - both SKILL.md texts, provenance-stripped, side by side
   - validate-plugin.sh output for the candidate
@@ -197,8 +209,14 @@ The critic may NOT read the builder's reasoning or this prompt's derivation.
 STATE
 Persist OUTSIDE the conversation, per skill: every attempt, every refutation, every
 T-metric, and the blind-guess outcome (guessed-correctly / could-not-tell). Round N must
-know what round N-1 already failed. Sink: a journal under the worktree, committed with
-the batch.
+know what round N-1 already failed.
+  SINK: a journal under the worktree, appended ATOMICALLY (write temp + rename) after
+  EACH event, BEFORE the next round starts. The batch commit reconciles what is already
+  on disk; it is not the moment the state first exists.
+  Why not "committed with the batch" alone: an interrupted batch would lose exactly the
+  round-to-round history the paragraph above requires — the sink would contradict its own
+  stated purpose. An earlier draft of this prompt said precisely that, and the review that
+  caught it named the contradiction rather than restating a preference.
 
 STOP  (every term below is measured, not adjectival)
 
--- a/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
+++ b/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
@@ -27,13 +27,16 @@
 | Check | skills | agents | commands |
 |---|---|---|---|
 | artifact file exists | ✅ | ✅ | ✅ |
-| frontmatter **parses as YAML** | ✅ | ⚠️ **top-level only** | ⚠️ **top-level only** |
+| frontmatter **parses as YAML** | ⚠️ **only if PyYAML present** | ⚠️ top-level only **+ PyYAML** | ⚠️ same |
 | `name` / `description` **present** | ✅ **hard fail** | ❌ only *warns* if frontmatter absent; keys never checked | ❌ same |
 | quality vs an external reference | ❌ | ❌ | ❌ |
 
 - Skills are the **strictest** class: `validate-plugin.sh:465` calls
   `scripts/validate-skill-frontmatter.sh`, which walks `rglob("SKILL.md")` and `sys.exit(1)` unless
   both `name` and `description` are present.
+- Without PyYAML the parse check `warn`s and the build can still PASS (`validate-plugin.sh:301`),
+  and `validate-skill-frontmatter.sh` uses `re.search`, not a parser — so "parses as YAML" is
+  **conditional on the runner**, not unconditional.
 - The YAML-parse gate globs `agents/*.md` / `commands/*.md`, and Python's `*` does **not** cross `/`
   — so **30 artifacts** (21 `agents/consultants/`, 9 `commands/*/`) have never been parsed. Filed
   as **#327**; measured blast of the fix is zero.
@@ -78,7 +81,7 @@ The 17 reference skills: `algorithmic-art` · `brand-guidelines` · `canvas-desi
 
 | Conjunct | Verdict | Basis |
 |---|---|---|
-| `bar_is_machine_verifiable` | **hybrid (not pure-D)** | **D** frontmatter parses · **T** description length, body word-count, disclosure-marker count · **J** blind A/B on clarity |
+| `bar_is_machine_verifiable` | **hybrid (not pure-D)** | **D** frontmatter parses · **T** description length, body word-count, linked-supporting-file count · **J** blind A/B on clarity |
 | `critic_is_independent_of_builder` | **partial — see below** | separate invocation without build history; the tag check is a *label* check, not a context check |
 | `evidence_is_directly_inspectable` | ✅ | both sides are readable files; the A/B is literal |
 | `long_loop_licensed` | **❌ FALSE** | the gate is an **AND** (`SKILL.md`): two conjuncts above are *hybrid* and *partial*, and partial is false in a conjunction ⇒ **the economic stop governs, not the profile budget** |
@@ -120,7 +123,7 @@ and its paired reference. One skill = one independently-judgeable piece.
 
 BUILD-METHOD
 A lead agent ranks the in-scope skills by measured gap (deterministic pass first:
-frontmatter fields, description length, body word-count, disclosure-marker count).
+frontmatter fields, description length, body word-count, linked-supporting-file count).
 Fan out one builder per skill; skills are independent — no verdict entanglement.
 
 Every builder spawn MUST route through the repository's delegation governance:
@@ -152,9 +155,16 @@ QUALITY-BAR  (named external artifacts — NOT adjectives)
                                                         no separate check needed here]
     T  description <= the spec's stated limit
     T  body <= 5,000 words
-    T  disclosure-markers >= 1 when body > 500 words, where a disclosure-marker is a
-       level-2 heading (`## `) or a `<details>` block. Both are countable; "progressive
-       disclosure is present" as a vibe is not.
+    T  progressive disclosure, per THIS REPO's definition (skills/skill-writer/SKILL.md
+       :184-195, :339 — "put advanced details in SEPARATE FILES" such as reference.md,
+       examples.md, scripts/, templates/, referenced from SKILL.md):
+         when body > 500 words, count sibling files in the skill directory that are
+         LINKED from SKILL.md. Require >= 1.
+       An earlier draft counted `## ` headings and <details> blocks instead. Neither
+       moves any material out of SKILL.md — a heading hides nothing, and <details> is
+       visual only since the agent receives the source regardless. That metric would
+       have passed large monolithic skills. It was invented without reading the
+       definition this repo already had.
     J  BLIND A/B, single success condition:
          the critic, shown the candidate and its paired reference with provenance
          stripped in randomised order, CANNOT reliably say which is Anthropic's.
@@ -202,9 +212,13 @@ STOP  (every term below is measured, not adjectival)
   EVIDENCE required for the exit: three journal entries, consecutive round numbers,
   each recording lens set, G(k)=0, and the critic's verbatim could-not-tell statement.
 
-  STAGNATION (early exit, failure): delta(k) = G(k-1) - G(k). Exit when delta(k) <= 0
-  for two consecutive rounds with G(k) > 0 — the critic keeps finding gaps and the
-  builder is not closing them. Report as NOT-CONVERGED; do not report as done.
+  STAGNATION (early exit, failure): measured on gap IDENTITY, not on the count.
+  A gap is identified by its verbatim recorded text. Exit NOT-CONVERGED when the SAME
+  gap is named in two consecutive rounds — the builder was handed it and did not close
+  it.
+  Counting instead would misfire on the common case: TASK asks each round to close the
+  LARGEST gap, so a builder closing one gap while the critic surfaces the next holds
+  G(k) constant at 1. A count-based delta reads that steady progress as stagnation.
 
   GOVERNING CAP: the ECONOMIC STOP, not the profile budget. long_loop_licensed is FALSE
   (see the gate table), so per SKILL.md the cap is convergence-engine's n*.
--- a/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
+++ b/skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md
@@ -22,14 +22,29 @@
 > rather than reading the claim — which is the only reason it was caught. Recorded because the
 > failure shape (an instrument whose *reach* differs from the claim's *scope*) recurs.
 
-**What is gated today, stated precisely** (`tests/validate-plugin.sh`, read not assumed):
+**What is gated today** — measured by reading `tests/validate-plugin.sh` *and the script it calls*:
 
 | Check | skills | agents | commands |
 |---|---|---|---|
 | artifact file exists | ✅ | ✅ | ✅ |
-| frontmatter **parses as YAML** | ✅ | ✅ | ✅ |
-| `name` / `description` **present** | ❌ **not checked** | ✅ | ✅ |
+| frontmatter **parses as YAML** | ✅ | ⚠️ **top-level only** | ⚠️ **top-level only** |
+| `name` / `description` **present** | ✅ **hard fail** | ❌ only *warns* if frontmatter absent; keys never checked | ❌ same |
 | quality vs an external reference | ❌ | ❌ | ❌ |
+
+- Skills are the **strictest** class: `validate-plugin.sh:465` calls
+  `scripts/validate-skill-frontmatter.sh`, which walks `rglob("SKILL.md")` and `sys.exit(1)` unless
+  both `name` and `description` are present.
+- The YAML-parse gate globs `agents/*.md` / `commands/*.md`, and Python's `*` does **not** cross `/`
+  — so **30 artifacts** (21 `agents/consultants/`, 9 `commands/*/`) have never been parsed. Filed
+  as **#327**; measured blast of the fix is zero.
+
+> ⚠️ **This table has now been wrong twice, in opposite directions.** The first render claimed
+> skills were unchecked for `name`/`description` and agents/commands were checked — almost exactly
+> backwards. The second render "corrected" it by reading only the top-level loops in
+> `validate-plugin.sh` and never following the call on line 465. Both times the instrument was a
+> partial read presented as a full one. The reviewer that caught it named the called script by path.
+
+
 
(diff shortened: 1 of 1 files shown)


ekson73:
@coderabbitai @qodo-code-review @chatgpt-codex-connector

All findings triaged against the code, not against the claim. Commit `8318a7d`.

**Accepted — and three of them were my own overstated claims, the same defect class each time: naming a mechanism without reading what it enforces.**

| # | Finding | Verification | Fix |
|---|---|---|---|
| 1 | CodeRabbit L13 — counts | Ran it: `git ls-files` gives **82/49/46 = 177**, not 82/28/37 = 147. `ls agents/*.md` stops at one directory level; git's pathspec `*` crosses `/`, so 21 consultant agents + 9 commands were invisible to my instrument, not absent. | Corrected, with the glob-reach failure recorded in the doc |
| 2 | qodo L12 — validator overstated | **Partly refuted, partly upheld.** YAML-parse **is** gated for skills (PR #322 added it; the check globs `skills/*/SKILL.md`). But `name`/`description` presence is gated for commands/agents **only**. I claimed both. | Replaced with a per-class table of what is actually checked |
| 3 | qodo L47 · codex L91 · CodeRabbit L49 — `convergence-guard` | **Upheld.** Read the script: it compares **caller-supplied** brand/axis tags, and its own help states `axis:author` ≠ `axis:reviewer` ⇒ ALLOW. A verifier inside the builder's conversation passes by picking a different label. It cannot observe context. | Downgraded to what it is (correlated-peer guard); independence moved to a procedural control; **residual named as unmitigated** |
| 4 | codex L82 — vacuous J-predicate | **Upheld, and the sharpest.** `cannot tell WHICH is Anthropic's` **OR** `names the gap` = `A OR NOT-A`. Every outcome passed, including a critic finding a clear gap. | `cannot-distinguish` is now the **only** success; any named gap is a refutation forcing another round |
| 5 | CodeRabbit L40 — unpinned bars | Upheld. | Pinned: repo commit `f6656c1256d5a8adfa37db9110046ef20bac644c` + spec `414,136 b` / sha256 `4022d0bd4b9b1d07…`; re-probe-and-record on drift |
| 6 | codex L73 — comparator leaks domain | Upheld. `skill-creator` as universal comparator lets the critic identify the reference by **subject matter**, measuring domain recognition rather than clarity. | Same-purpose pairing; no same-purpose reference ⇒ out of scope, recorded |
| 7 | codex L68 — fan-out bypasses GaaS | Upheld. | Routes through `skills/delegate-governance` + `plugin-scripts/gaac/delegate.sh init\|dna\|finalize`; both verified present + executable |
| 8 | codex L79 — undefined threshold | Upheld. | Countable: **≥ 1** `## ` heading or `<details>` block when body > **500** words |
| 9 | qodo L49 — `long_loop_licensed` ambiguous | Upheld. | Split into `long_loop_licensed` (policy ✅) vs `run_armed` (this run ❌) |
| 10 | CodeRabbit L5 — missing signature | Upheld. | `Claude-Doc-ef60-325` + ISO 8601 timestamps added |
| 11 | qodo L69 — broken `[C04]` | Upheld. | Spelled out: *Git Worktree Protocol ([C04], `~/.claude/CLAUDE.md`)* |

**Nothing refused outright**; #2 is the only partial, and its refuted half is documented above rather than silently dropped.

Every new numeric claim was self-verified before commit — `82/49/46/177` re-derived from `git ls-files`, and both governance paths tested present + executable. `validate-plugin.sh` rc=0 captured **unpiped**; `gitleaks` clean.

Re-review please.

ekson73:
@coderabbitai @chatgpt-codex-connector @qodo-code-review

Round 2. All four findings **upheld** after verification — and one of them exposed a real defect in this repo, not just in the document. Commit `bf5befe`.

| # | Finding | Verification | Fix |
|---|---|---|---|
| 1 | codex L31 — validator matrix inverted | **Upheld.** Followed the call this doc had never followed: `validate-plugin.sh:465` invokes `scripts/validate-skill-frontmatter.sh`, which walks `rglob("SKILL.md")` and `sys.exit(1)` unless `name` **and** `description` are present. Skills are the **strictest** class. The agent/command loops only `warn` when frontmatter is **absent** and never check the keys. | Matrix corrected; the fact that it was wrong **twice, in opposite directions** is recorded in the document rather than quietly fixed |
| 2 | *(consequence of verifying #1)* | The YAML gate added in #322 globs `agents/*.md`; Python's `*` does not cross `/`. **30 artifacts** (21 `agents/consultants/`, 9 `commands/*/`) have never been parsed. | Filed as **#327**, blast measured **zero** — see below |
| 3 | codex L165 — "no new gap" | **Upheld, and it contradicted my own acceptance rule two sections above.** A critic re-naming the *same* unresolved gap three times passes a novelty test while the skill stays distinguishable. | Exit is now three **consecutive** rounds with `G(k) = 0` |
| 4 | coderabbit L169 + codex L167 — stop-rule terms undefined | **Upheld.** `no new gap`, `delta`, `threshold`, `lens set` were adjectives. | Gap = one critic-named distinguishing feature, recorded verbatim · `G(k)` = per-round count · lens set = `k mod 6` over a fixed order, so the three exit rounds provably use three distinct lenses · exit evidence = three consecutive journal entries · stagnation = `delta(k) = G(k-1) − G(k) ≤ 0` twice with `G(k) > 0`, reported **NOT-CONVERGED** rather than done. No third operator slot needed — the metric is mechanical |

### The finding worth naming separately

Verifying #1 surfaced that **PR #322's gate has the exact defect it was built to fix, one level up.** Its own commit message argued:

> *a grep answers "does the text contain?", never "does the parser accept?"*

It then asked *"does the parser accept?"* for the files it could see, and never asked *"do I see all the files?"* — `glob.glob("agents/*.md")` = 28, `git ls-files 'agents/*.md'` = 49.

Blast radius of the fix is **measured zero**: 24 of the 30 uncovered files parse cleanly, and the other 6 are frontmatter-less fragments the gate already skips via its existing `continue`.

**Filed as #327 rather than fixed inline**, deliberately: `tests/validate-plugin.sh` is a validation gate, and whether that counts as a guardrail surface is genuinely ambiguous. The rule says fail-closed on uncertainty, which would make this docs PR ratification-gated for a reason unrelated to its content. Recorded rather than dropped.

### Self-verification before commit

`465` still calls the script · the script still uses `rglob` + `sys.exit(1)` · the agent/command loops still only `warn` · the reach gap is still exactly **30** · #327 is `OPEN`. `validate-plugin.sh` rc=0 captured **unpiped**; `gitleaks` clean.

@coderabbitai your `CHANGES_REQUESTED` is from `b6995e7`; three commits have landed since. Re-review please.

ekson73:
@chatgpt-codex-connector @coderabbitai @qodo-code-review

Round 3. Both findings **upheld**. Commit `5caf692`.

**1 · The gate table contradicted itself (L84, P1).** The governing predicate in `SKILL.md` is literally an `AND` of three:

```
long_loop_licensed := bar_is_machine_verifiable
                  AND critic_is_independent_of_builder
                  AND evidence_is_directly_inspectable
```

The table marked two of those three `hybrid` and `partial` — then marked the conclusion that depends on them ✅, three rows below. **Partial is false in a conjunction.**

The consequence is not cosmetic: it selects a **different stopping rule**. `long_loop_licensed = FALSE` ⇒ the convergence-engine **economic stop** governs, not the profile budget the STOP section was using. And since ρ is not estimable for a prose-clarity loop, the corrected STOP cites **no round count** — it uses convergence-engine's **parameter-free escape** (harness cap + stagnation test), which is precisely the branch #324 populated in that SSOT this morning for exactly this case.

**2 · The prompt carried the error the matrix had just fixed (L144, P2).** Round 2 corrected the analysis table and left the *rendered prompt* telling its runner that skill `name`/`description` are *"NOT gated today — check it directly"*, when `validate-plugin.sh:465` hard-fails on exactly that. Fixing the analysis and shipping the stale executable text hands the contradiction to whoever runs it. Removed.

---

**Running tally across three rounds — every one of these was mine:**

| Round | Defect | Shape |
|---|---|---|
| 1 | counts 147 vs 177 | instrument reach ≠ claim scope (`ls` vs git pathspec) |
| 1 | `convergence-guard` "enforces fresh context" | named a mechanism without reading what it enforces |
| 1 | J-predicate `A OR NOT-A` | acceptance rule that accepted everything |
| 2 | validator matrix inverted | partial read presented as a full one (never followed the call on line 465) |
| 2 | *(consequence)* #322's gate has the defect it was built to fix | asked "does the parser accept?", never "do I see all the files?" → **#327** |
| 3 | gate ✅ over two failed conjuncts | contradiction carried in plain sight, and it changed the stopping rule |

Four bots found six real defects in a 143-line document about *not asserting things you have not measured*. The document is better for it, and that tally is now recorded **inside** the artifact rather than only in this thread.

`validate-plugin.sh` rc=0 **unpiped** · `gitleaks` clean · every round self-verified before commit.

@coderabbitai your `CHANGES_REQUESTED` is still pinned to `b6995e7` — four commits back. Re-review when you can.

ekson73:
@coderabbitai review

ekson73:
@chatgpt-codex-connector @coderabbitai @qodo-code-review

Round 4. All three upheld. Commit `d94e408`.

| Finding | Verification | Fix |
|---|---|---|
| **L157** progressive disclosure | **Upheld.** The metric counted `## ` headings and `<details>`. Neither moves material out of `SKILL.md` — a heading hides nothing, `<details>` is visual only since the agent gets the source anyway. It would have passed large monolithic skills. **This repo already had the definition**: `skills/skill-writer/SKILL.md:184-195, :339` — *"put advanced details in SEPARATE FILES"*. | Count sibling files in the skill dir **linked from** `SKILL.md`; require ≥1 when body > 500 words |
| **L149** YAML gate conditional | **Upheld.** Without PyYAML, `validate-plugin.sh:301` **warns** and the build can still PASS; `validate-skill-frontmatter.sh` uses `re.search`, not a parser. | Matrix now marks the parse gate **conditional on the runner** |
| **L207** stagnation misfires | **Upheld, and on the *common* case.** `TASK` asks each round to close the **largest** gap — so a builder closing one while the critic surfaces the next holds `G(k)` constant at 1, and a count-based delta reads sustained progress as stagnation. | Measured on gap **identity**: exit when the *same verbatim gap* is named twice consecutively |

I also swept two stale echoes of the redefined term (`disclosure-marker` in the gate table and `BUILD-METHOD`) — leaving them would have repeated round 3's own L144 defect: correcting the analysis and shipping the old wording in the executable text.

---

**L157 is the fourth instance of one shape in this PR**: naming something without reading its definition. `convergence-guard` (round 1), the validator matrix (round 2), the `AND` predicate (round 3), progressive disclosure (round 4). Each time the definition was *already in this repo* and each time a reviewer found it by reading the source instead of the claim.

**Nine defects across four rounds, every one of them mine, in a document whose entire subject is not asserting what you have not measured.** That is either an indictment or the best possible demonstration that the review loop works. I think it is the second, and the tally now lives inside the artifact rather than only in this thread.

`validate-plugin.sh` rc=0 **unpiped** · `gitleaks` clean · every round self-verified before commit.

@coderabbitai — your `CHANGES_REQUESTED` is pinned to `b6995e7`, **five commits back**, and an explicit `@coderabbitai review` did not dislodge it. I will **not** dismiss my own blocking review to unblock my own PR; that is the self-approval seam. Holding for you.

ekson73:
**Round 5 (round n+6 recon session).** CodeRabbit's one open finding — `10 attempts/piece` retained as a profile-derived limit after stating the profile budget does not govern — **upheld and fixed** (`90de1a5`):

- `HARNESS CAP` line: ATTEMPTS becomes `operator-set` like COST CEILING and WALL-CLOCK; **no default rendered** (a number there would be invented — same defect class round 4 retracted). Refusal rule extended to *any*-unset-cap.
- Synced the §Budget quotation of the STOP rule (`either cap` → `ANY cap`) the first commit left stale.

Force-push note: `d6b15e9→90de1a5` amends my own just-pushed commit (no third-party fetch between); feature branch, not protected.

ekson73:
@coderabbitai

Your 9 findings were scored against the **current** head (`f0fe65f`), not the head you reviewed (`b6995e7`, now 7 commits back). Reporting honestly: 6 were already addressed in rounds 2–5, 3 were still actionable and are fixed now, and one of those three is a **declared non-fix** rather than a fix.

| # | Finding | Disposition | Commit |
|---|---|---|---|
| L13 | Update the corpus inventory | ✅ fixed — 177 (82/49/46); your count was right and mine was wrong. Credited in-file: the glob `ls agents/*.md` stops at one level, git's pathspec `*` crosses `/`. You ran the count instead of reading the claim, which is the only reason it was caught. | `8318a7d` |
| L40 | Pin the external quality references | ✅ fixed — `anthropics/skills @ f6656c12…` + spec 414,136 b sha256 `4022d0bd…` | `bf5befe` |
| L69 | Resolve `[C04]` before fan-out | ✅ fixed — resolved by path + all spawns routed through `delegate-governance` | `bf5befe` |
| L74 | Task-matched references for blind A/B | ✅ fixed — PAIRING block by domain proximity; no same-purpose reference ⇒ candidate is OUT OF SCOPE, not paired badly | `5caf692` |
| L112 | Define the stop-rule measurements | ✅ fixed — GAP defined as the unit, `G(k)` per-round count, thresholds named | `5caf692` |
| L52 | Make the D/T/J gate executable | ✅ fixed — each level now carries a countable procedure (incl. `VERIFY FIRST: python3 -c "import yaml"`) | `d94e408` `90fb20b` |
| **L103** | **Checkpoint the journal before batch completion** | ✅ **fixed** — and it was the strongest finding of the nine: it exposed a contradiction *inside one paragraph*. STATE demanded *"Round N must know what round N-1 already failed"* and then set the sink to *"committed with the batch"* — an interrupted batch loses exactly that history. Now an atomic per-event append before the next round; the batch commit reconciles. | `f0fe65f` |
| **L49** | **Make critic isolation runner-enforced** | ⚠️ **declared non-fix, with rationale in-file.** It cannot be enforced here: `bin/convergence-guard` compares **caller-supplied** labels, so a caller that mislabels a context-contaminated critic passes the guard. Isolation is a prompt-level contract. Recorded as a known hole + what would close it (a runner that spawns the critic and owns its context), rather than prose implying a gate exists. | `f0fe65f` |
| **L5** | **Add the required document signature** | ✅ fixed at the root. The agent ID is there. The hand-written `revised` stamp I added went stale **three times in one session, silently**, because it duplicates a fact git already tracks — so it now points at `git log -1 --format=%cI` instead of copying it. | `f0fe65f` |

`tests/validate-plugin.sh` rc=0 (captured unpiped — a pipe would have reported `tail`'s status) · gitleaks: no leaks.

Note on process: your verdict is 7 commits old, and this repo's own `pr-reviewer-communication.md` allows proceeding after a 30-minute bot silence. I did not take that route — measuring the findings against the current head found one still valid (L103), so the age of the verdict turned out to be uncorrelated with the validity of its contents. Could you re-review `f0fe65f`?

ekson73:
**Round 7 (this pi session).** The ATTEMPTS-slot inconsistency flagged on `f0fe65f` is closed via stacked PR **#332** (merged `2026-08-15T00:17:36Z`, squash `45531b5`): §Budget now lists **three** required slots — `<ATTEMPTS>`, `<COST CEILING>`, `<WALL-CLOCK>` — matching the ANY-cap refusal rule. Root of the residue was my own round-5 commit (`90de1a5`); closing it is that commit's responsibility (no-loose-endings). Ownership of this PR stays with the parallel session — requesting re-review: @coderabbitai review

ekson73:
@codex All six findings on `90fb20b`/`f0fe65f` are addressed in `281ab03`. The first P1 was the one that mattered — thank you for it.

| # | Finding | Fix |
|---|---|---|
| **P1 L189** | Keep candidate-only validation out of the blind critic input | **Fixed, and the leak was worse than asymmetry.** I ran it: the script's FIRST output line is the banner `Multi-Agent OS Plugin Validation` followed by an absolute path containing `multi-agent-os`, and each `pass()` prints the candidate's own skill name. The critic would not have *inferred* provenance — it would have **read** it. Your option (a), running it on both sides, does not work either: the reference is not a MAOS plugin, so the validator condemns it structurally for reasons unrelated to quality — a signal that always condemns one side is a second leak, not symmetry. So: **D-level checks move to a LEAD-run PRE-GATE** (a FAIL bounces to the builder and never reaches the critic), and the critic sees only the two provenance-stripped texts + T-metrics. That is *already* what the LENS ASSIGNMENT clause said the permitted inputs were — the document had been disagreeing with itself, and removing the dump made it consistent rather than poorer. |
| **P1 L231** | Require clean coverage of all six lenses before exiting | **Fixed.** `G=0 for three consecutive rounds` over a six-lens rotation covers half the lenses, exactly as you describe. Worse, the draft cited *"the three exit rounds provably use three distinct lenses"* as a **strength** — three distinct is three **missing**. EXIT is now: every one of the six lenses returned `G=0` on its **most recent** run. Self-anchoring (a lens that found a gap must be re-run clean), and the consequence is stated plainly: minimum six rounds, since you cannot call a skill indistinguishable on an axis you never tested. |
| P2 L174 | Make the progressive-disclosure check reachable | **Fixed** — and it was the second failure of the same check. Draft 1 counted headings/`<details>` (wrong proxy, moves nothing out of the file); draft 2 gated on `body > 500 LINES` while the metric above requires `<= 500 LINES` (**dead code** — unreachable on any passing candidate, so a 500-line monolith with zero supporting files passed untested). Both let a monolith through. The threshold is now the **paired reference's own** linked-sibling count, measured on both sides with no gate — non-invented and always reachable. |
| P2 L153 | Keep unmatched skills within the declared target | **Fixed by narrowing the promise, not the rule.** TARGET is now the *pairable subset*; the pool is 82 but the subset size is unknown until PAIRING runs, so it is not stated. The unpairable remainder is an **output** — enumerated, counted, reported with each domain. A blind A/B against a reference that does not exist is not a weaker measurement; it is not a measurement. |
| P2 L258 | Track persistent gaps with stable identities | **Fixed.** Verbatim identity loses to paraphrase + lens rotation, as you say. The **LEAD** now assigns a stable ID on first record (skill + lens + slug) and marks each open gap closed-or-open by **inspecting the artifact**, never by diffing critic prose. Identity is a decision about the artifact, not a comparison of two sentences about it. |
| P2 L294 | Include the attempts cap in the arming form | **Fixed** — concurrently by a parallel session (`45531b5`, PR #332) and by me; I merged both annotations rather than picking one. An operator filling the printed form exactly would have been refused for a field it was never asked for: a fill-in form that cannot satisfy its own gate is worse than none, because it reads as completable. |

`tests/validate-plugin.sh` rc=0 (unpiped, re-run after the rebase) · gitleaks: no leaks.

Pattern worth naming across rounds 5-7: every decisive finding has been an **internal contradiction**, not a preference — the sink that contradicted the history it required, the evidence list that defeated the blindness it demanded, the exit that proved its own hole, the check gated out of reach by the metric above it. Each read fine in isolation. Reviewing the sentence against its neighbours is what caught them.

ekson73:
**Merge record (standing auto-merge authorization, `--auto-merge` default disposition — gates below all PASS).**

| Gate | Verdict |
|---|---|
| Checks @ `45531b5` | **12 SUCCESS** (governance-validation · release-coherence · self-test+44-matrix · gitleaks · Trivy ×2 · snyk ×2 · pip-audit · OpenSSF · Lockfile · Amazon Q); Scorecard skipping = neutral |
| Agentic convergence | CodeRabbit: **"Re-review passed on current head 45531b5 — no further blocking inconsistency"** (comment `5299439xxx` chain). The formal `CHANGES_REQUESTED` state is pinned to `b6995e7`, 9 commits back — staleness of VERDICT ≠ staleness of FINDINGS; all 9 scored + fixed rounds 2–7 (L103 journal-atomicity, L49 declared-non-fix with rationale, L5 git-native signature, ATTEMPTS slot via #332) |
| Guardrail surface | diff = 1 docs file (`skills/refine-braindump-to-prompt/examples/GAUNTLET-maos-skill-corpus.md`), additive, **0 hits** |
| HUMAN_DOMAIN / ⛔ | none — docs artifact, loop NOT armed (budget slots remain operator-set by design) |

Cross-session note: this session contributed rounds 5 + 7 (the round-5 residue was closed by its own author, per no-loose-endings); the owning session's rounds 1–6 did the rest. The artifact stays **unarmed** — arming requires the operator's three caps, as designed.

PR #333, opened 2026-08-15
fix(tests): the YAML gate reached only top-level — 30 artifacts never parsed

### LINKED ISSUE #327: validate-plugin.sh: the YAML gate reaches only top-level — 30 artifacts never parsed
## The gate has the same defect it was built to fix, one level up

PR #322 added a YAML-parse gate to `tests/validate-plugin.sh`, with this rationale in its own commit message:

> *a grep answers "does the text contain?", never "does the parser accept?"; the second question is the one that decides whether the artifact loads at all.*

The gate is real and it works. But its **reach** was never measured — and Python's `glob.glob("agents/*.md")` does **not** cross `/`, unlike git's pathspec of the same shape.

### Measured (2026-08-14, repo root)

```
glob('agents/*.md'    ) = 28      git ls-files 'agents/*.md'   = 49
glob('commands/*.md'  ) = 37      git ls-files 'commands/*.md' = 46
```

**30 artifacts have never passed the YAML gate** — 21 under `agents/consultants/` and 9 under `commands/*/`.

So the fix answered *"does the parser accept?"* for the files it could see, and never asked *"do I see all the files?"* Same failure shape, one level up.

### Blast radius of the fix: measured ZERO

Of the 30 uncovered files:
- **24** have frontmatter and **parse cleanly**
- **6** (`commands/auto-shard/operations/*.md`) have no frontmatter at all, and the gate's existing `if not t.startswith("---\n"): continue` already skips those

Making the glob recursive therefore covers 30 more artifacts and introduces **no new failures**.

### Proposed fix

In the `YAML_BAD` heredoc in `tests/validate-plugin.sh`:

```python
for pat in ("skills/*/SKILL.md", "agents/*.md", "commands/*.md"):
    for p in sorted(glob.glob(os.path.join(root, pat))):
```
→
```python
for pat in ("skills/**/SKILL.md", "agents/**/*.md", "commands/**/*.md"):
    for p in sorted(glob.glob(os.path.join(root, pat), recursive=True)):
```

Plus a **reach assertion** so this cannot regress silently: compare the glob's file count against `git ls-files` for the same patterns and fail on divergence. A gate that cannot prove its own coverage is a gate that can shrink without anyone noticing.

### Why this is filed rather than fixed inline

Found while writing #325 (a docs artifact describing what the validator gates). `tests/validate-plugin.sh` is a validation gate, and whether that counts as a guardrail surface is genuinely ambiguous — the rule says fail-closed on uncertainty, which routes it to operator ratification. Mixing it into a docs PR would have made that PR ratification-gated for a reason unrelated to its content.

Recorded rather than silently dropped, per no-silent-drop.

### Also worth deciding

`commands/auto-shard/operations/*.md` are frontmatter-less fragments included by a parent command, not standalone artifacts — much like `README.md`, which the gate already excludes by name. Consider excluding them explicitly instead of relying on the `continue`, so "skipped because it is a fragment" is distinguishable from "skipped because it is broken".

ekson73:
Closes #327

## What

`tests/validate-plugin.sh`'s YAML-parse gate globbed `agents/*.md` and `commands/*.md` **non-recursively**. Python's `*` does not cross `/` — unlike git's pathspec of the same shape — so **30 artifacts had never been parsed** while every run reported *"all artifact frontmatter parses as YAML"*.

That is the same failure shape as the defect the gate was built to fix, one level up. PR #322's own rationale was *"a grep answers 'does the text contain?', never 'does the parser accept?'"* — correct, and then its **reach** was never measured.

## Changes

| Change | Why |
|---|---|
| `**` patterns **+ `recursive=True`** | Both are load-bearing: `**` *without* the flag behaves as plain `*`, so a half-fix would have looked right and changed nothing. |
| **REACH ASSERTION** | `git ls-files` is the ground truth for what *should* have been parsed. The gate now **fails** if it ever under-reaches again. Degrade-safe: no git / not a repo → skipped, because this validator also runs from installed plugin dirs with no `.git`. |

The assertion is the root fix. Without it, this exact class recurs silently — it has already bitten three times in one session (an artifact count, a validator-coverage table, and this gate).

## Re-measured on current `main`

The counts in #327 had already moved (`transmute` added a skill and a command), so they were re-derived rather than copied:

```
skills/*/SKILL.md      83  →  83   (unchanged — skills are always one level deep)
agents/*.md            28  →  49   (+21, agents/consultants/)
commands/*.md          38  →  47   (+9,  commands/*/)
                      149  → 179   reach gain: 30
```

## Verification — controls, not inspection

| Control | Result |
|---|---|
| **Blast** | `validate-plugin.sh` rc=0 (unpiped), *"all artifact frontmatter parses as YAML"*, *"Plugin is ready for use!"* |
| **Positive** | recursive glob reaches 30 more files (149 → 179) |
| **Negative** | revert the glob, keep the assertion → `REACH GAP :: 30 tracked artifact(s)…` fires |
| **git-absent** | complete tree minus `.git` → rc=0, zero false positives; **identical to `origin/main` on the same tree** (one variable varied) |

⚠️ **The negative control caught a defect in the assertion itself.** `seen.add()` initially sat *after* the `continue` that skips files without frontmatter — so it recorded *"parsed"* while the assertion claimed *"reached"*, and reported **7 deliberately-skipped files as a reach gap**. A reach assertion with a reach defect. Moved to the first line of the loop body; it now reports the true **30**.

An earlier run of that same degrade test also gave a false `rc=1` — caused by my copying an *incomplete* tree, not by the fix. Re-run with one variable varied (same tree, `origin/main`'s script vs mine) it was identical. Both are recorded because the session's recurring lesson is precisely that an instrument's reach must be proven, not assumed.

## Risk

Docs/test-tooling only; no runtime path. Blast measured zero. Fully revertible.

### DIFF
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -276,8 +276,18 @@ if python3 -c "import yaml" 2>/dev/null; then
 import glob, io, os, sys, yaml
 root = sys.argv[1]
 bad = []
-for pat in ("skills/*/SKILL.md", "agents/*.md", "commands/*.md"):
-    for p in sorted(glob.glob(os.path.join(root, pat))):
+seen = set()
+# `**` + recursive=True: Python's `*` does NOT cross `/`, unlike git's pathspec of the
+# same shape. With the non-recursive patterns this gate never saw agents/consultants/*.md
+# nor commands/*/*.md — 30 artifacts that were reported as covered. See issue #327.
+# NOTE both halves are load-bearing: `**` without recursive=True behaves as plain `*`.
+for pat in ("skills/**/SKILL.md", "agents/**/*.md", "commands/**/*.md"):
+    for p in sorted(glob.glob(os.path.join(root, pat), recursive=True)):
+        seen.add(os.path.relpath(p, root))   # REACHED — must precede every `continue`
+                                             # below, or `seen` records "parsed" while the
+                                             # assertion claims "reached" (they differ for
+                                             # files with no frontmatter, which are skipped
+                                             # on purpose and are NOT a reach gap).
         if os.path.basename(p) == "README.md":
             continue                      # docs, not artifacts (same rule as above)
         try:
@@ -289,6 +299,27 @@ for pat in ("skills/*/SKILL.md", "agents/*.md", "commands/*.md"):
             bad.append(f"{os.path.relpath(p, root)} :: {str(e).splitlines()[0]}")
         except Exception as e:
             bad.append(f"{os.path.relpath(p, root)} :: {type(e).__name__}: {e}")
+
+# REACH ASSERTION — the gate must not silently under-reach again (issue #327).
+# git's pathspec `*` crosses `/`, so these patterns are the ground truth for what
+# SHOULD have been parsed. Degrade-safe: no git / not a repo -> skip, never fail,
+# because this validator also runs from an installed plugin dir with no .git.
+try:
+    import subprocess
+    tracked = set()
+    for pat in ("skills/*/SKILL.md", "agents/*.md", "commands/*.md"):
+        out = subprocess.run(["git", "-C", root, "ls-files", pat],
+                             capture_output=True, text=True, timeout=10)
+        if out.returncode != 0:
+            raise RuntimeError("git unavailable")
+        tracked.update(f for f in out.stdout.split() if os.path.basename(f) != "README.md")
+    missed = sorted(tracked - seen)
+    if missed:
+        bad.append(f"REACH GAP :: {len(missed)} tracked artifact(s) the glob never parsed, "
+                   f"e.g. {missed[0]} — the gate under-reaches (see #327)")
+except Exception:
+    pass                                  # no git here; the glob fix above still applies
+
 print("\n".join(bad))
 PYEOF
 )
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -331,7 +331,10 @@ if os.path.exists(os.path.join(root, ".git")):
                                  capture_output=True, text=True, timeout=10)
             if out.returncode != 0:
                 raise OSError(f"git ls-files rc={out.returncode}")
-            tracked.update(f.replace("\\", "/") for f in out.stdout.split()
+            # splitlines(): git ls-files is newline-delimited. `.split()` would corrupt any
+            # tracked path containing spaces (zero today — probed — but the gate's whole
+            # purpose is to not silently under-report again, per #327).
+            tracked.update(f.replace("\\", "/") for f in out.stdout.splitlines()
                            if os.path.basename(f) != "README.md")
         missed = sorted(tracked - seen)
         if missed:
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -283,7 +283,14 @@ seen = set()
 # NOTE both halves are load-bearing: `**` without recursive=True behaves as plain `*`.
 for pat in ("skills/**/SKILL.md", "agents/**/*.md", "commands/**/*.md"):
     for p in sorted(glob.glob(os.path.join(root, pat), recursive=True)):
-        seen.add(os.path.relpath(p, root))   # REACHED — must precede every `continue`
+        # REACHED — must precede every `continue` below (see note under the except).
+        # `.replace(os.sep, "/")`: on Windows os.path.relpath returns `agents\foo.md`
+        # while `git ls-files` ALWAYS emits `/`, so without this every artifact lands in
+        # `tracked - seen` and the assertion reports a false total reach gap on every run.
+        # os.sep (not a literal "\\"): on POSIX os.sep == "/" so this is a no-op, which
+        # matters because `\` is a legal filename character there and a literal replace
+        # would corrupt it.
+        seen.add(os.path.relpath(p, root).replace(os.sep, "/"))
                                              # below, or `seen` records "parsed" while the
                                              # assertion claims "reached" (they differ for
                                              # files with no frontmatter, which are skipped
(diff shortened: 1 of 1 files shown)


ekson73:
@codex Fixed in `bb1053c` — and you were right for a reason I want to record, because it is the same class as the bug this PR fixes.

**The finding.** On Windows (Git Bash + native Python), `os.path.relpath()` returns `agents\foo.md` while `git ls-files` **always** emits `/`. So `tracked - seen` contains every artifact → a false **total** reach gap on every run → a Windows contributor cannot pass the repo's required validation. Confirmed by simulation:

```
relpath(Windows) = {'agents\consultants\x.md'}
git ls-files     = {'agents/consultants/x.md'}
without fix: tracked - seen = {'agents/consultants/x.md'}   ← false REACH GAP
with fix:    tracked - seen = set()
```

**`os.sep`, not a literal `"\\"`** — your wording said "normalize both sets to one separator", and the choice of *which* normalization matters: on POSIX `\` is a legal filename character, so a literal `.replace("\\", "/")` corrupts it. Measured:

```
'agents/we\ird.md'.replace(os.sep, '/')  → 'agents/we\ird.md'   ✅ preserved (os.sep == '/' → no-op)
'agents/we\ird.md'.replace('\\',   '/')  → 'agents/we/ird.md'   ❌ corrupted
```

**Why this happened, recorded in the commit.** The parent commit shipped with three controls — blast, positive, negative — and all three ran on macOS. They proved cross-platform correctness the test environment never reached. That is precisely the defect this PR exists to fix (an instrument whose *reach* is narrower than the claim's *scope*), one layer up: not in the glob, but in the test of the glob. You caught it by reasoning about a platform the tests could not visit.

Regression: `validate-plugin.sh` rc=0 (unpiped) · negative control still reports the true `REACH GAP :: 30` · gitleaks clean.

ekson73:
@qodo-code-review Both findings addressed in `46ba7ec`. The first one was sharp — it caught the guard disabling itself.

| # | Finding | Disposition |
|---|---|---|
| **L328** | Exceção ampla oculta asserção | **Fixed.** `except Exception: pass` meant any unexpected error silently switched off the control that exists to prevent under-reach. Now `.git` absent is the **one** silent skip (installed plugin dir — expected and documented); `.git` present + git failing is an **anomaly** and is reported as `REACH ASSERTION DID NOT RUN :: <type>`. Excepts narrowed to `(OSError, subprocess.SubprocessError)` so a real bug propagates instead of being swallowed. |
| **L289** | Separador de path inconsistente | **Already fixed** in `bb1053c`, and your suggested line is verbatim what was applied — you and codex converged on it independently. I also took your optional half (normalizing `tracked` defensively), so both sides are POSIX-separated regardless of platform. |

**Your L328 fix required an order I did not expect.** Narrowing the excepts alone would have been a *regression*: a propagating exception exits python non-zero with **empty stdout**, and the shell tested `[ -z "$YAML_BAD" ]` → "nothing bad found" → **PASS**. So the exit-code check had to land first.

⚠️ **And my first attempt at that check was itself dead code** — worth recording, because it is the same defect class this PR exists to fix. I "proved" the false-pass with `OUT=$(python3 -c 'raise ...')` in a plain shell. But this script runs under `set -euo pipefail` (line 9), where the failing assignment **aborts before the next line** — so `YAML_RC=$?` never executed and my message never fired. The probe did not run under the conditions it claimed to describe. Fixed with `|| YAML_RC=$?`, which keeps the failure local so it can be *reported* instead of ending the script wordlessly.

It was caught only by asking *"did it fail for the **right** reason?"* instead of accepting `rc=1`. An exit code that is correct by accident looks identical to one that is correct by design.

Five paths, all run against the real script under its own set-options:

| Path | Result |
|---|---|
| normal | rc=0, zero errors |
| python crash | rc=1 + `✗ frontmatter YAML check crashed (python rc=1) — result is UNKNOWN, not clean` |
| `.git` absent | rc=0, silent skip |
| `.git` present + broken git | `REACH ASSERTION DID NOT RUN :: OSError` |
| negative control | `REACH GAP :: 30 tracked` |

gitleaks clean. Could you re-review `46ba7ec`?

ekson73:
@coderabbitai review — head is now `46ba7ec`. Qodo's 3 findings disposition: #1 broad-except FIXED (46ba7ec: .git-gate + narrow except + loud UNVERIFIED report) · #2 separator FIXED (bb1053c: os.sep→/ normalize) · #3 `.split()` space-corruption: measured exposure in this repo = **0 tracked artifact paths contain spaces** (`git ls-files` probe); latent-pattern issue, theoretical-only here.

ekson73:
@chatgpt-codex-connector

P2 (inherited `YAML_RC`) — **confirmed and fixed** in `198c64b`. I reproduced it before touching anything rather than taking the finding on faith:

```
# before
YAML_RC=7 bash tests/validate-plugin.sh
  -> rc=1   ✗ frontmatter YAML check crashed (python rc=7)
```

Clean tree, Python exited 0, and the gate reported a crash — from a value it never measured. You are right that `${YAML_RC:-0}` does not cover this: it substitutes on *unset/null*, not on *set-to-garbage*. Those are different failure modes and I had conflated them.

Fix is a scope-local init before the capture (`YAML_RC=0`), and the now-redundant `${YAML_RC:-0}` line is dropped — one mechanism, not two.

| path | result |
|---|---|
| A clean, `YAML_RC` unset | `rc=0`, 0 crash msgs |
| B clean, `YAML_RC=7` inherited | `rc=0`, 0 crash msgs *(was `rc=1`)* |
| C python crash **+** `YAML_RC=7` | `rc=1` → `crashed (python rc=1)` — the **measured** value, not the inherited one |
| D narrowed-glob negative control | `REACH GAP :: 30 tracked` still fires |

**C** is the one that matters: it proves the reset did not blind the guard — it made it report what it observed instead of what it inherited.

Direction note: this failed *closed* (false alarm), not open. Lower blast radius than the silent-bypass class — but a validator that cries wolf on a clean tree trains people to ignore it, which converts fail-closed into fail-open at the culture layer. Worth the line.

This is the fifth instance today of one shape: **the instrument's reach or state exceeding what the claim was entitled to assert.** The glob that did not cross `/`; the `seen` set that recorded *parsed* while claiming *reached*; tests run only on macOS claiming cross-platform; a probe run outside `set -e` claiming to describe the script; and now a variable read from outside its own scope. Thanks for catching the one I authored.

Issue #336, opened 2026-08-15
validate-plugin.sh: the SHELL frontmatter loops still don't cross '/' — 21 agents + 9 commands unchecked


ekson73:
Sibling of #327, **still live on main** (`a499524`). #333 fixed the *Python YAML gate*; the *shell frontmatter loops* 130 lines above it have the identical `*`-doesn't-cross-`/` defect.

## Measured

```
agents/*.md      -> 28      git ls-files 'agents/*.md'    -> 49   (21 unchecked)
commands/*.md    -> 38      git ls-files 'commands/*.md'  -> 47   ( 9 unchecked)
agents/consultants/alexandr-wang.md: tracked=1, mentions in validator output=0
```

## Fixture test — with positive control

| case | result |
|---|---|
| **top-level** `agents/toplevel-broken.md`, no frontmatter | `⚠ toplevel-broken.md missing frontmatter` ✅ **caught** |
| **nested** `agents/consultants/broken.md`, no frontmatter | zero mentions ❌ **not caught** |
| **nested** `commands/sub/broken.md`, no frontmatter | zero mentions ❌ **not caught**, and count reported `No commands found` |

The top-level control proves the check works — the nested miss is a real coverage hole, not a broken probe.

## Impact

- An artifact with **no frontmatter at all** in `agents/consultants/` or `commands/*/` passes validation silently. (The #333 Python gate catches *broken YAML*; it does not substitute for the *presence* check.)
- `AGENT_COUNT` / `COMMAND_COUNT` under-report — the validator states a coverage number smaller than reality while printing `✓ PASSED`.

## Lines

`tests/validate-plugin.sh:140` (`for cmd in "$PLUGIN_ROOT/commands"/*.md`) · `:166` (`for agent in "$PLUGIN_ROOT/agents"/*.md`) · `:117` (`skills/*/` — verify whether nested skills exist before changing).

## Note

This is the 8th instance of one class on this file today (#327/#333 glob · seen-after-continue · os.sep · set-e dead probe · inherited YAML_RC · split/splitlines #334 · NUL+fsdecode #335 · this). Root fix already shipped in #333 — the **reach assertion** — but it only guards the Python set. Consider extending it to assert the shell loops' coverage too, so the twin can't drift again.

ekson73:
Fixed by **#337** (merged `dd95c50`): find-based recursion + relative-path messages + dir guards; Qodo re-review Bugs (0); controls positive/negative/missing-dir all green.

[...]

PR #337, opened 2026-08-15
fix(tests): shell frontmatter loops recurse via find — close #336

ekson73:
Closes #336.

## [PR-as-Documentation-Prompt]

**Context.** Sibling of #327, still live on main: #333/#335 fixed the *Python YAML gate*; the *shell presence loops* ~130 lines above had the identical `*`-doesn't-cross-`/` defect — 21 nested agents (`agents/consultants/`) + 9 nested commands never received the missing-frontmatter check, and their absence corrupted the counts (`No commands found` when only nested exist).

**Objective.** Same reach discipline as #333: recurse the presence check.

### Change (`tests/validate-plugin.sh`, 2 loops)
- `"$PLUGIN_ROOT/{commands,agents}"/*.md` glob loop → `find -type f -name '*.md' | sort` via **process substitution** (`while IFS= read -r`).
- **Why find, not globstar**: `shopt -s globstar` requires bash 4+; macOS stock bash is 3.2 — the fix would itself fail on the operator's machine. `find` is POSIX and version-free.
- **Why process substitution, not pipe**: `find | while` forks a subshell and the count is lost (the loop's own `((COUNT++))` would silently never reach the summary — the same silent-under-report class this issue is about).
- README.md exclusion, warn/pass semantics, count logic: unchanged.

### Verification (controls, not claims)
| Control | Result |
|---|---|
| Positive: nested no-frontmatter fixtures planted | `⚠ tmp-broken-336.md missing frontmatter` + `⚠ sub336.md missing frontmatter` — **caught** (were invisible before) |
| Counts with fixtures | 49 agents / 47 commands = the issue's own `git ls-files` numbers |
| Negative: fixtures removed | 48 / 46 (Δ1 = README.md exclusion, pre-existing) · full validator **rc=0** |

guardrail-surface: 0 hits (CI test script, not hook/rule/allowlist).

PR #335, opened 2026-08-15
fix(tests): NUL-delimited ls-files + fsdecode — CodeRabbit's #333 finding, carried forward

ekson73:
## [PR-as-Documentation-Prompt]

**Context.** CodeRabbit's full review of #333 @ `265ad99` (CHANGES_REQUESTED, **1 actionable, Minor**) landed ~3 min before #333 merged — the finding never closed: tracked-file collection must preserve paths exactly.

**Objective.** Close it against main: `git ls-files -z`, byte-split on NUL, `os.fsdecode()`, retain git's forward slashes, drop `replace("\\\\", "/")` (corrupts legal POSIX filenames containing a backslash). Supersedes the `splitlines()` step; same contract, exact now.

### Verification
- `tests/validate-plugin.sh` rc=0 — the reach assertion itself is the A/B (runs the `-z` path vs the glob, reports no gap).
- guardrail-surface: 0 hits (CI test script) · +8/−6, one file.

PR #338, opened 2026-08-15
fix(tests): harden #337 — NUL-safe walk, POSIX head, and a shell reach assertion

ekson73:
> **Re-scoped.** A peer session solved #336 in **#337**, which is **merged** (`dd95c50`). I rebuilt this branch **on top of** that commit rather than re-landing work already in `main`. What remains is the residual delta — three things #337 does not have. **#336 is closed by #337, not by this.**

## Comparison that drove the re-scope

| | #337 (merged) | this |
|---|---|---|
| loops recurse | ✅ | inherited |
| NUL-safe (newline in path) | ❌ `find \| sort`, line-based | ✅ `-print0` + `read -d ''` |
| **shell reach assertion** | ❌ (the `REACH GAP` on main is the *Python* gate, `:362`) | ✅ new |
| POSIX `head -n 1` | ❌ `head -1` | ✅ |
| `sort` portability | ✅ plain `sort` — **better than my first attempt** | ✅ probed fallback |

#337 is better than my original on one axis (plain `sort` beats my `sort -z`); that is now moot since `sort -z` here is probed with a `cat` fallback.

## The three additions

**1 — NUL-safe walk.** A newline is a legal POSIX filename character; a line-based read turns one such path into two bogus entries (`2 real paths → 3 entries`, measured). This is the exact residual `-z` closed on the *Python* side in #335 — the shell half still had it.

**2 — `head -n 1`.** POSIX specifies `head -n number`; `head -1` is a legacy BSD extension. qodo finding on #338.

**3 — Shell reach assertion (the root fix).** #337 fixed *the reach*; this proves it **stays** fixed — the visited set is compared against `git ls-files` and the run **fails** on a gap. Sibling of the Python assertion from #333. The surface defect was the glob; the root is that *the loops could not say what they had not seen*.

## Two defects in the assertion itself

Both caught by negative controls printing **nothing** — neither by any clean run:

| defect | consequence |
|---|---|
| `[ -d .git ]` is **false** in a linked worktree (`.git` is a *file*) | assertion skipped **silently** where the work happens |
| gating on `git rev-parse` collapsed *"not a repo"* and *"git broke"* into one non-zero exit | silent skip **either way** — error-as-absence |

Resolved with `-e "$PLUGIN_ROOT/.git"`, the predicate the Python half already used. An assertion that cannot run where the work happens is worse than none: it reports ✓ while measuring nothing.

## Controls

| | case | result |
|---|---|---|
| A | clean | `rc=0` · `reached every tracked artifact (177)` |
| B | agents `-maxdepth 1` | `REACH GAP :: 21 tracked` |
| C | commands `-maxdepth 1` | `REACH GAP :: 9 tracked` |
| D | nested no-frontmatter | 2 caught (`agents/` + `commands/`) |
| E | non-repo fixture | 0 false alarms |
| F | `git` rc=3 | 2 × `DID NOT RUN` (shell + Python halves) |
| G | `sort -z` unavailable | `reached every tracked artifact (177)` |

gitleaks clean. Diff vs `main`: **+66 / −6**, purely additive.

## qodo L116 (JIRA key) — refuted

`ekson73/multi-agent-os` is a personal/community repo with **GitHub Issues enabled**; this PR carries `Refs #336`/`#337`. Routing here is **personal → Linear · Vek → Jira · community/repo-scoped → GitHub Issue**. Jira is the *Vek corporate* tracker — a `VKS-####` key on a personal-repo PR would point at a work item that does not exist and would cross a boundary kept deliberately separate. Traceability gate is met, by the artifact this repo actually uses.

Refs #336 (closed by #337) · #337 · #327 · #333 · #335.

### DIFF
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -161,6 +161,20 @@ if [ -d "$PLUGIN_ROOT/commands" ]; then
 # `-print0` + `read -d ''` rather than line-based: a NEWLINE is a legal POSIX filename
 # character, and a line-based read turns one such path into two bogus entries. Same
 # residual `-z` closed on the Python side in #335. `head -n 1` is the POSIX spelling.
+#
+# Discovery runs in the PARENT (into a file) instead of `done < <(find ...)`: a process
+# substitution runs the producer in a SEPARATE process, so a `find` that dies reaches the
+# loop as plain end-of-input — indistinguishable from "the dir was empty". Neither `set -e`
+# nor `pipefail` sees it, and the run reports GREEN having validated nothing (CodeRabbit
+# #338, Major). As a parent statement, `pipefail` makes `$?` reflect ANY stage and
+# `|| RC=$?` captures it without aborting; the loop then reads a plain file, so the
+# process-substitution class is gone rather than patched.
+CMD_LIST="$(mktemp -t mao-cmdlist.XXXXXX)"
+CMD_DISCOVER_RC=0
+find "$PLUGIN_ROOT/commands" -type f -name '*.md' -print0 | $SORT_NUL > "$CMD_LIST" || CMD_DISCOVER_RC=$?
+if [ "$CMD_DISCOVER_RC" -ne 0 ]; then
+    fail "command discovery FAILED (rc=$CMD_DISCOVER_RC) — coverage is UNVERIFIED, not clean"
+fi
 while IFS= read -r -d '' cmd; do
     if [ -f "$cmd" ]; then
         cmd_name=$(basename "$cmd")
@@ -176,7 +190,8 @@ while IFS= read -r -d '' cmd; do
             ((COMMAND_COUNT++)) || true
         fi
     fi
-done < <(find "$PLUGIN_ROOT/commands" -type f -name '*.md' -print0 | $SORT_NUL)
+done < "$CMD_LIST"
+rm -f "$CMD_LIST"
 fi
 
 if [ $COMMAND_COUNT -eq 0 ]; then
@@ -194,7 +209,14 @@ AGENT_COUNT=0
 # `agents/*.md` could not — 21 consultants were invisible to this presence check.
 # Relative-path messages + dir guard, same rationale as the commands loop.
 if [ -d "$PLUGIN_ROOT/agents" ]; then
-# Same `-print0` / `head -n 1` rationale as the commands loop above.
+# Same `-print0` / `head -n 1` rationale as the commands loop above, and the same
+# parent-statement discovery so a failing `find` cannot masquerade as an empty dir.
+AGENT_LIST="$(mktemp -t mao-agentlist.XXXXXX)"
+AGENT_DISCOVER_RC=0
+find "$PLUGIN_ROOT/agents" -type f -name '*.md' -print0 | $SORT_NUL > "$AGENT_LIST" || AGENT_DISCOVER_RC=$?
+if [ "$AGENT_DISCOVER_RC" -ne 0 ]; then
+    fail "agent discovery FAILED (rc=$AGENT_DISCOVER_RC) — coverage is UNVERIFIED, not clean"
+fi
 while IFS= read -r -d '' agent; do
     if [ -f "$agent" ]; then
         agent_name=$(basename "$agent")
@@ -210,7 +232,8 @@ while IFS= read -r -d '' agent; do
             ((AGENT_COUNT++)) || true
         fi
     fi
-done < <(find "$PLUGIN_ROOT/agents" -type f -name '*.md' -print0 | $SORT_NUL)
+done < "$AGENT_LIST"
+rm -f "$AGENT_LIST"
 fi
 
 if [ $AGENT_COUNT -eq 0 ]; then
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -110,6 +110,17 @@ for script in "session-start.sh" "pre-delegate.sh" "post-delegate.sh" "session-e
 done
 echo ""
 
+# Every artifact the loops below actually visit is recorded here (repo-relative) so the
+# reach assertion at the end can PROVE the walk did not under-reach. #337 fixed the
+# reach; this makes a future regression impossible to land silently. See #336.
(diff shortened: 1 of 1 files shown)


ekson73:
@qodo-code-review

Two accepted and fixed in `5881a6f`; one refuted with rationale.

| # | Finding | Verdict |
|---|---|---|
| L241 | `head -1` not POSIX | ✅ **accepted** — fixed at 3 sites |
| L166 | `sort -z` not portable | ✅ **accepted** — now probed, not assumed |
| L116 | missing JIRA key | ❌ **refuted** — wrong tracker for this repo |

### L241 — `head -1` → `head -n 1`

Correct: POSIX specifies `head -n number`; the short form is a legacy BSD extension. Fixed at all three sites. Worth noting for the record that **two were carried forward** from `main` and **the third was genuinely mine** (the `REACH GAP` message I added in `9716b61`) — so the "código novo" framing was accurate for the one that mattered.

### L166 — `sort -z`

Accepted, though the diagnosis needed one correction: `sort -z` **does** work here (measured: `printf 'b\0a\0' | sort -z` → ok on this macOS), and if it were absent the failure would be **loud, not silent** — the stream would empty, the loops would visit nothing, and the new reach assertion would fire `REACH GAP :: 177`. So it was not a stability risk in the way described.

It was still worth fixing, for a simpler reason: `sort -z` exists here **purely to stabilize output order** — no check depends on it. Carrying a portability constraint for a cosmetic benefit is a bad trade. Now probed at startup: available → use it; absent → `cat`.

**New control G** (shadow `sort` with one that rejects `-z`): `rc=0`, 46 commands, 48 agents, `reached every tracked artifact (177)`.

### L116 — JIRA key: refuted

This repo is **`ekson73/multi-agent-os`** — personal/community, **GitHub Issues enabled**, and this PR carries **`Closes #336`**. That *is* the traceability link, in the tracker that governs this repo.

Our routing rule is explicit that the tracker depends on the world: **personal → Linear · Vek → Jira · community/repo-scoped → GitHub Issue**. Jira is the **Vek corporate** tracker. Adding a `VKS-####` key to a personal-repo PR would not add traceability — it would point at a work item that does not exist, and it would cross a boundary we keep deliberately separate (a "Two-Worlds" violation: Vek and personal identities/repos/trackers never mix).

So the traceability gate is **met**, just not by the artifact the check looked for. The finding appears to apply a corporate-repo ruleset to a community repo. No change made.

---

Re-verified on `5881a6f`: `A` clean `rc=0`/177 · `B` agents `-maxdepth 1` → `REACH GAP :: 21` · `F` git `rc=3` → `DID NOT RUN` (both shell and Python halves) · `G` no-`sort -z` → 177 · gitleaks clean.

Could you please re-review?

ekson73:
@coderabbitai

**Finding accepted as valid — and already resolved on the current head, by inheritance rather than by me.** Details, because "already fixed" deserves proof rather than assertion:

Your review targets `5881a6f` (`original_commit_id` confirms). That commit **did** have the bug:

```
5881a6f:159   cmd_rel="${cmd#$PLUGIN_ROOT/}"      <- unquoted, your finding
5881a6f:190   agent_rel="${agent#$PLUGIN_ROOT/}"  <- unquoted
```

Reproduced exactly as you describe:

```
root='/tmp/plug[ab]'  path='/tmp/plug[ab]/commands/x.md'
  unquoted -> /tmp/plug[ab]/commands/x.md   (stays ABSOLUTE — pattern didn't match)
  quoted   -> commands/x.md
```

Current head `c21cdfe`:

```
c21cdfe:167   cmd_rel="${cmd#"$PLUGIN_ROOT"/}"
c21cdfe:201   agent_rel="${agent#"$PLUGIN_ROOT"/}"
```

**Timing**: your review landed `02:08:02Z`; `c21cdfe` was pushed `02:08:55Z` — 53 seconds later. You reviewed the head that existed at the time; nothing wrong on your side.

**The honest part**: this was *not* fixed because I spotted it. This PR was re-scoped mid-flight — a peer session solved the same underlying issue in #337, which merged, so I `reset --hard`ed onto `dd95c50` and re-applied only my residual delta. #337's author had **already quoted it correctly**; the bug was mine and the rebase inherited their fix. Credit where it belongs.

Requesting a re-review of `c21cdfe` so the stale `CHANGES_REQUESTED` can clear — the current diff is +66/−6 on top of the merged #337.

@coderabbitai review

ekson73:
@coderabbitai

Both findings addressed on `5ee1b38`. I verified each against the current head rather than trusting the SHA stamp.

| Finding | Verdict | Evidence |
|---|---|---|
| L159/L190 — unquoted `${var#$PREFIX/}` (SC2295, Minor) | **already fixed** on `c21cdfe` | negative control: 2 unquoted occurrences on the flagged `5881a6f`, **0** on `c21cdfe` (2 properly quoted). Resolved by inheriting the quoting from #337 — the peer had it right, the bug was mine. |
| L170 — producer failure not propagated (Major) | **live and real — fixed now** | measured below |

**On the Major one — you were right, and my reach assertion did not cover it.**

`done < <(find ...)` runs the producer in a separate process, so a failing `find` arrives as plain end-of-input, indistinguishable from an empty directory. The reach assertion I added catches that *inside a git checkout*; outside one (installed plugin, tarball, CI without `.git`) it is skipped entirely and the failure is invisible.

Measured on a non-git tree, varying one thing (sabotaged `find` flag), same tree, same script otherwise:

```
OLD  sabotage OFF   rc=0   48 commands validated   0 fails
OLD  sabotage ON    rc=0    2 commands validated   0 fails   <- 46 silently skipped, GREEN
NEW  sabotage ON    rc=1    2 commands validated   "command discovery FAILED (rc=1) — coverage is UNVERIFIED, not clean"
```

The fix eliminates the class instead of patching it: discovery runs as a **parent statement** into a file (`pipefail` makes `$?` reflect any stage; `|| RC=$?` captures it without aborting), then the loop reads a plain file — no process substitution remains in either loop.

Positive control on a healthy tree: 234 pass / 0 fail / 7 warn, reach assertion still green at 177 artifacts.

This is the 7th defect the review bots caught in this PR, and the most valuable — a validator that can silently under-report is the worst possible instrument, which is precisely what this PR exists to fix. Could you please re-review `5ee1b38`?

PR #341, opened 2026-08-15
fix(tests): classify entries vs docs — the recursion was right, the classification was not

### LINKED ISSUE #339: layout: 7 validator warnings are true signals — a SKILL.md under commands/, and a policy doc registered as an agent
#337 made the shell loops recurse, which surfaced **7 `⚠ missing frontmatter` warnings that were previously invisible**. They are **not** validator false positives — they point at two real layout facts. Filing rather than deciding: the remedy is a layout call, not a test call.

## Measured

**(a) `commands/auto-shard/` is skill-shaped but lives under `commands/`**

```
commands/auto-shard/
  SKILL.md          <- has frontmatter; it is a SKILL, not a command
  operations/*.md   <- 6 files, no frontmatter, titled "ShardAnalyzer Operation" etc.
  scripts/
```

- The 6 `operations/*.md` are **internal modules of that skill**, not invocable commands. `grep -rl 'auto-shard/operations'` finds no reference to them as commands outside their own directory.
- The validator now demands command frontmatter from them because they are under `commands/`.
- Separately: `commands/auto-shard/SKILL.md` is reached by **neither** skill check — the shell loop globs `skills/*/SKILL.md` and the Python gate globs `skills/**/SKILL.md`. Both are rooted at `skills/`. A `SKILL.md` under `commands/` is validated by nothing.

**(b) `agents/COWORK-AUTONOMY-POLICY.md` is a policy document in the agents namespace**

First line: `# Cowork Autonomy Policy — ≥0.90-No-HITL`. Not an agent definition. It is nonetheless **registered by the loader as an agent** — it appears in the available agent list as `maos:COWORK-AUTONOMY-POLICY`, described generically as "Agent from maos plugin", because it has no frontmatter to describe itself.

## Why this is worth a decision rather than a silent suppression

The warnings are the validator working correctly — a namespace is being used for two different kinds of thing. Suppressing them (path exclusions) would restore the silence #327/#336 were about removing.

## Options (not choosing — layout is not a test concern)

| | option | note |
|---|---|---|
| A | move `commands/auto-shard/` → `skills/auto-shard/` | matches its actual shape; makes its `SKILL.md` validated; `operations/` stop being command-shaped |
| B | move `agents/COWORK-AUTONOMY-POLICY.md` → `docs/` or `protocols/` | stops the loader registering a policy as an agent |
| C | add frontmatter to all 7 | cheapest, but declares them as things they are not |
| D | teach the validator that `commands/*/operations/` is skill-internal | narrowest, but encodes one directory's convention into the gate |

A+B look right to me; C would make the validator green by lying to it. Deferring to you because moving files in `commands/` and `agents/` changes what the plugin loader registers, which is a blast-radius question, not a linting one.

Surfaced by #337/#338. Refs #336, #327.

ekson73:
Closes #339.

## What

The recursion landed in #333/#337/#338 was correct. Its side-effect was not: the loops began reaching files that were **never commands** — 6 sub-documents of a skill (`commands/auto-shard/operations/*.md`) and a policy SSOT (`agents/COWORK-AUTONOMY-POLICY.md`) — and warned `missing frontmatter` about artifacts that were never supposed to have any.

**A validator that warns about non-defects trains people to ignore it** — converting fail-closed into fail-open at the cultural layer, in the very file built to be trustworthy.

## Why not simply move the files

I measured both before deciding, and they are **opposite** cases:

| Artifact | External refs | State | Call |
|---|---|---|---|
| `commands/auto-shard/` | **0** | has `SKILL.md` + `operations/` + `scripts/`; discovered by **neither** loader path — dead weight from a bulk consolidation | genuinely misfiled, but relocating it **activates a dormant, never-reviewed skill** in a published plugin → left to a separate decision |
| `agents/COWORK-AUTONOMY-POLICY.md` | **10** (red-team · ooda-loop · praxis-audit · gap-loop · protocols · CHANGELOG) | live, heavily-referenced SSOT | **correctly** in `agents/` — it is policy *for* agents; moving it breaks 10 refs |

So the root is not layout, it is **classification**: the validator treated every `.md` under `commands/` as a command.

## How

**ONE** predicate, `is_entry()`, consumed by the loops **and** by the reach assertion's expectation. Two notions of *what counts* is exactly how the sets drift apart (the #336 class, one level up).

- **(a)** ALL-CAPS stem ⇒ doc by repo convention (`README`, `SKILL`, `COWORK-AUTONOMY-POLICY`) — generalizes the former literal `!= README.md`, which knew about only one of them
- **(b)** anything under a directory holding `SKILL.md` is that skill's territory

Deliberately **not** "skip anything nested": `commands/code/analyze/dependencies.md` is a genuinely nested command and keeps being validated.

## Two traps inside the predicate itself, both measured

1. **Test the stem, not the basename.** `.md` is lowercase by definition, so a lowercase test against `README.md` always matches on the extension and the rule never fires.
2. **`[[:lower:]]`, not `[a-z]`.** The bracket *range* is resolved by locale **collation**, which interleaves `aAbBcC…` in any UTF-8 locale — so `[a-z]` **matches `README` under bash**. It does *not* under zsh.

   My isolated truth-table ran under zsh (the interactive shell) while the script runs under bash (shebang). Same code, opposite verdict: **a green unit test for code that was broken in the shell that actually executes it.** `[[:lower:]]` agrees across bash/zsh × C/en_US/pt_BR (6/6 measured).

## Reach assertion is now bidirectional

Filtering both sides through one predicate kills drift — but makes a one-directional check **vacuous**: a predicate that skips everything shrinks both sets and they agree trivially. Measured: the assertion went **green on a deliberately broken predicate**.

- `MISSED` (tracked ∖ visited) → the loops under-reach
- `EXTRA` (visited ∖ tracked) → the expectation shrank below what was visited

## Controls

| Control | rc | Detection |
|---|---|---|
| POSITIVE — healthy tree | 0 | 233 pass / **0 warn** / 0 fail (was 234/7/0); reach 169 entries, 10 classified as docs |
| NEG — loops under-reach (`-maxdepth 1`) | 1 | `REACH GAP` |
| NEG — producer dies (bogus `find` flag) | 1 | `discovery FAILED` (#338 guard intact) |
| NEG — predicate drops skills | 1 | `EXPECTATION GAP :: 83 artifact(s)` |

## Honest residual — documented in-file, not claimed fixed

A predicate that drops **commands/agents symmetrically** still evades both directions, because `is_entry` filters both sides. It is surfaced by **reporting the skip count** in the pass line (10 → 170 is visible to a reader), not by a gate. A guard-for-the-guard-for-the-guard is where this stops being proportionate.

## Out of scope (deliberately)

Relocating `commands/auto-shard/` → `skills/`. It is dead today; moving it *activates* unreviewed surface in a published plugin. That is a real decision, not a cleanup, and it is orthogonal to making the validator honest.

### DIFF
--- a/tests/validate-plugin.sh
+++ b/tests/validate-plugin.sh
@@ -320,7 +320,8 @@ if [ -e "$PLUGIN_ROOT/.git" ]; then
         is_entry "$tracked_path" && printf '%s\n' "$tracked_path" >> "$TRACKED_LIST"
     done < "$TRACKED_RAW"
     TRACKED_RAW_N=$(wc -l < "$TRACKED_RAW" | tr -d ' ')
-    rm -f "$TRACKED_RAW"
+    # TRACKED_RAW stays alive past the comparison: EXTRA must be able to ask "is this path
+    # git-tracked at all?" to tell real drift from a not-yet-committed file (qodo #341).
     if [ "$REACH_RC" -ne 0 ]; then
         fail "reach assertion DID NOT RUN (git ls-files rc=$REACH_RC) — coverage is UNVERIFIED, not clean"
     else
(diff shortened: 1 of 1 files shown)


ekson73:
@qodo-code-review

Finding accepted and fixed on `babfcaa` — **reproduced first**, not taken on faith.

## The finding was right

```
$ cat > commands/zz-probe.md <<'EOF'   # eligible, uncommitted
...
$ bash tests/validate-plugin.sh .
rc=1  ✗ EXPECTATION GAP :: 1 artifact(s) ... — is_entry over-skips
```

Wrong diagnosis for the right observation: the file is not dropped by `is_entry`, it simply is not committed yet. And it breaks the normal dev loop — write a command, validate, *then* commit — which contradicts why the walk is filesystem-driven at all: **git is the independent cross-check, not the source of truth.** I introduced that regression together with the bidirectional check in the same PR.

## Fix — your patch direction, adopted

`EXTRA` is split by tracked-ness **before** the verdict chain (`TRACKED_RAW` kept alive past the comparison, as you suggested):

| `EXTRA` member | Meaning | Verdict |
|---|---|---|
| present in `TRACKED_RAW` | git knows it, yet the expectation dropped it | **real drift** → `fail` |
| absent from git | not committed yet | counted + reported, **never fails** |

## Controls — all four re-run, not just the one being fixed

| Control | rc | Result |
|---|---|---|
| POSITIVE healthy | 0 | 233 pass / 0 warn / 0 fail |
| **your case** — untracked command | **0** | `169 entries; 10 classified as docs/sub-documents; 1 not yet committed` |
| NEG predicate drops skills | 1 | `EXPECTATION GAP` still fires |
| NEG loops under-reach | 1 | `REACH GAP` still fires |
| NEG producer dies | 1 | `discovery FAILED` still fires |

Re-running the **whole** matrix earned its keep: my first attempt at this fix replaced the `elif` with a standalone `if`, leaving `if [ -n "$MISSED" ]` unclosed. The script died with `unexpected end of file` and *every* control returned rc=2 — invisible if I had only checked the case I was fixing.

Also on this PR: @amazon-q-developer reviewed `f203da4` with no findings. Could you re-review `babfcaa`?

